PR #1, opened 2026-04-09
Add per-user token budget management

nvasilevv:
## Summary

- **Per-user token budgets** managed exclusively by platform admins — users are metered on all OpenAI-powered features (chat, compaction, brain summarization) while background pipeline tasks (Gemini extraction, embeddings, screener) are excluded
- **Pre-flight budget check** returns HTTP 429 when a user's budget is exhausted; `budget=0` means unlimited so existing users aren't blocked before admin configures budgets
- **Audit trail** via `TokenUsageLog` table with per-request breakdown by model, operation, and chat — admin can view recent usage logs per user
- **Admin UI** shows token budget/usage column with color-coded progress bar (green/amber/red) and a "Set Budget" modal supporting both absolute set and incremental add modes

### Backend changes
| File | Change |
|------|--------|
| `backend/app/models/token_usage.py` | New `TokenUsageLog` model |
| `backend/app/models/user.py` | Added `token_budget`, `tokens_used` fields |
| `backend/app/db/init_db.py` | Model registration + migration SQL |
| `backend/app/services/token_budget.py` | `check_budget()` + `record_usage()` with atomic SQL increment |
| `backend/app/services/chat_stream.py` | `stream_options={"include_usage": True}`, usage capture, `token-usage` SSE events |
| `backend/app/api/routes.py` | Pre-flight check, usage accumulation across agentic steps, recording |
| `backend/app/services/brain.py` | Record brain summarization usage |
| `backend/app/api/chats.py` | Record compaction endpoint usage |
| `backend/app/api/admin.py` | `PATCH /admin/users/{id}/token-budget`, `GET /admin/users/{id}/token-usage`, updated `UserResponse` |

### Frontend changes
| File | Change |
|------|--------|
| `frontend/src/lib/types.ts` | `AdminUser` extended with budget fields |
| `frontend/src/lib/api.ts` | `adminSetTokenBudget()`, `adminGetTokenUsage()` |
| `frontend/src/components/admin/user-management.tsx` | Token Budget column with progress bar + Set Budget modal |

## Test plan

- [ ] Start backend, verify migration runs (`token_budget` and `tokens_used` columns exist on `user` table)
- [ ] As admin, set a token budget via `PATCH /api/admin/users/{id}/token-budget`
- [ ] As that user, send a chat message — verify `tokenusagelog` row created and `user.tokens_used` increments
- [ ] Exhaust the budget — verify next chat returns 429
- [ ] As admin, add more tokens — verify user can chat again
- [ ] Verify admin UI shows budget/usage columns with progress bar and "Set Budget" modal works
- [ ] Verify `budget=0` users (no budget set) are not blocked


nvasilevv:
## PR #1 Review: "Add per-user token budget management"

### Summary
This PR adds an admin-controlled per-user token budget system that meters OpenAI API usage across chat, compaction, and brain summarization operations. It introduces a `TokenUsageLog` audit table, budget check/enforcement with HTTP 429, atomic SQL increments for usage tracking, and an admin UI with progress bars and a budget-setting modal. The PR description is **accurate and comprehensive** — it matches the actual code changes well.

---

### Review Findings

**🔴 Critical/Security**

1. **Race condition on budget check vs. usage recording** (`routes.py:1048-1054`, `token_budget.py:64-72`). The pre-flight `check_budget()` reads the user's remaining tokens, but usage is only recorded *after* the streaming response completes. Between the check and the record, another concurrent request could pass the check too. This is a classic TOCTOU (time-of-check-time-of-use) race. For a budget system the impact is overspend, not data loss, so severity depends on how strict budgets need to be — but it should be documented or mitigated.

2. **`add_tokens` on the admin PATCH endpoint uses ORM-level increment** (`admin.py:83`): `user.token_budget += body.add_tokens`. If two admins PATCH simultaneously, one increment could be lost. The `record_usage` function correctly uses `SET tokens_used = tokens_used + :delta` (atomic SQL), but the admin budget setter does not. Should use the same atomic SQL pattern.

**🟡 Improvement**

3. **Token usage log joins on `user_email` (string) rather than `user_id` (UUID)** (`token_usage.py:11`). If a user's email ever changes, historical logs become orphaned. Using `user_id` as the FK would be more resilient and consistent with the `chat_id` FK pattern already used in the same model.

4. **`_compact_messages` return type change is a breaking interface change** (`chat_stream.py:102-107`). It now returns a tuple `(str, dict | None)` instead of `str`. All callers were updated, but the function lacks a type annotation for the return — adding `-> tuple[str, dict | None]` would prevent future callers from missing the usage dict.

5. **No budget check before compaction or brain summarization** (`chats.py:369`, `brain.py:68`). The pre-flight check only gates the chat endpoint (`routes.py`). A user at budget limit could still trigger compaction (which fires automatically during chat) and brain summarization (background task). This means tokens continue to be consumed and *recorded* even after the budget is "exhausted." The budget check should arguably wrap these too, or at minimum this should be a documented design choice.

6. **Hardcoded model string in brain.py** (`brain.py:79`): `model="gpt-4.1-mini"` is a string literal. If the actual model used by the brain call changes, this log entry will be wrong. Should reference the same constant/variable used in the API call.

7. **Bare `except Exception` with silent swallow** (`chats.py:183`, `brain.py:81`, `routes.py:246`). Usage recording failures are silently logged as warnings. This is reasonable for resilience, but means budget tracking can silently drift. Consider at minimum logging the exception message (not just "Failed to record...").

**🔵 Nitpick**

8. **`tokens_remaining` computed in multiple places** (`admin.py:33,43,98,127`): `max(0, user.token_budget - user.tokens_used)` is repeated 4 times. Could be a property on the `User` model or a helper.

9. **Inline import of `TokenUsageLog`** (`admin.py:109`): `from app.models.token_usage import TokenUsageLog` is inside the function body. The other model imports in this file are at the top level. Same pattern in `brain.py:69` and `routes.py:1048`. While avoiding circular imports is a valid reason, these modules don't have circular dependency risks — consider moving to top-level.

10. **`adminGetTokenUsage` is defined but never called** (`api.ts:728-749`). The function exists in the API client but isn't used anywhere in the frontend. If it's for future use, fine, but it's dead code today.

---

### Positive Feedback

- **Atomic SQL increment** in `record_usage` (`token_budget.py:58-64`) is the right approach — avoids read-modify-write races on the hot path.
- **`budget=0` means unlimited** is a pragmatic default that avoids breaking existing users.
- **SSE `token-usage` events** that are captured in the stream producer but *not* forwarded to the frontend is clean separation — metering stays server-side.
- **The PR description is excellent** — table of changes per file, clear test plan, accurate summary.
- **Progress bar color logic** (green/amber/red) in the admin UI is a nice touch for quick visual scanning.

---

### Conclusion

**Request Changes** — the atomic budget increment issue (#2) is a straightforward fix (use raw SQL like `record_usage` does), and the TOCTOU race (#1) should at minimum be documented as a known limitation. The `user_email` vs `user_id` FK choice (#3) is worth discussing before this ships, as it's harder to change later. The rest are improvements that could be addressed in follow-ups.

nvasilevv:
Minor nit on the updated `brain.py` (lines 411-416):

```python
async with get_session_context() as usage_session:
    brain_user = (await usage_session.exec(
        select(User).where(User.email == user_email)
    )).first()
if brain_user:
    await record_usage(...)
```

The `if brain_user:` and `record_usage()` call sit outside the `async with` block. This works fine in practice — `brain_user.id` is a plain UUID already loaded into memory, and `record_usage` opens its own session — but the indentation makes it look like an accidental dedent. Moving `if brain_user:` inside the `async with` would make the intent clearer.

Not a blocker — just a readability thing.

nvasilevv:
LGTM — all review findings addressed. Clean implementation.

PR #3, opened 2026-04-09
Add encryption at rest for sensitive database fields

nvasilevv:
## Summary

- **New**: Application-level encryption at rest using Fernet (AES-128-CBC + HMAC-SHA256) for sensitive database fields. Encryption key never touches the database server — all crypto happens in the Python application layer.
- **New**: Enables `pgcrypto` PostgreSQL extension as a foundation for future database-level encryption needs.
- **New**: Idempotent backfill migration that encrypts existing plaintext data on startup (500 rows per batch per table).

### Encrypted fields

| Table | Field | Method |
|---|---|---|
| `chatmessage` | `content` | Fernet |
| `chatmessage` | `parts_json` | Fernet |
| `chat` | `compacted_context` | Fernet |
| `screenerrule` | `webhook_url` | Fernet |
| `screenerrule` | `email_recipients` (each item) | Fernet |
| `user` | `password_reset_token` | HMAC-SHA256 hash |

### Files changed

- `backend/app/core/encryption.py` — New module: `encrypt()`, `decrypt()`, `encrypt_json_list()`, `decrypt_json_list()`, `hmac_hash()`
- `backend/app/core/config.py` — Added `db_encryption_key` setting
- `backend/app/api/routes.py` — Encrypt chat messages on write, decrypt screener fields on read, encrypt screener fields on create/update
- `backend/app/api/chats.py` — Decrypt messages on load, encrypt compacted context on save
- `backend/app/api/auth.py` — HMAC-hash password reset tokens on store and lookup
- `backend/app/services/screener.py` — Decrypt webhook URL before delivery
- `backend/app/services/screener_email.py` — Decrypt email recipients before sending
- `backend/app/db/init_db.py` — Enable pgcrypto extension + backfill migration
- `backend/pyproject.toml` — Added `cryptography>=44.0.0` dependency
- `deploy/write_env.py` — Added `DB_ENCRYPTION_KEY` to required env vars
- `.github/workflows/deploy-hetzner.yml` — Passes `DB_ENCRYPTION_KEY` secret

### Design decisions

- **Fernet over pgcrypto column encryption**: Columns stay `TEXT` (no schema changes), queries remain normal SQLModel/SQLAlchemy, encryption key never sent to the DB server.
- **Backwards compatible**: Unencrypted values (no `enc::` prefix) pass through unchanged, so the migration is gradual and safe.
- **Password reset tokens use HMAC**: We only need to verify, not decrypt — deterministic hashing allows WHERE-clause lookups without storing the original token.
- **No encryption on queryable fields**: `user_email`, `chat_id`, timestamps, and other filter/join columns remain plaintext.

## Pre-deploy checklist

- [x] `DB_ENCRYPTION_KEY` secret added to GitHub repo
- [ ] Verify migration runs cleanly on staging/prod (check logs for `Encrypted N rows` messages)
- [ ] If >500 rows exist in any table, confirm subsequent restarts pick up remaining rows

## Test plan

- [ ] Create a new chat and verify messages are stored encrypted in DB (`SELECT content FROM chatmessage` should show `enc::...` prefix)
- [ ] Load an existing chat and verify messages decrypt correctly in the UI
- [ ] Trigger a password reset flow and confirm token is hashed in DB and reset link works
- [ ] Create/update a screener rule with webhook URL and verify it's encrypted in DB
- [ ] Trigger a screener and verify webhook delivery and email delivery still work with encrypted fields
- [ ] Verify compaction still works (compact a chat, check `compacted_context` is encrypted)


nvasilevv:
## PR Review — Encryption at Rest

### Summary
Solid PR — the `enc::` prefix scheme, HMAC for reset tokens, and keeping the key out of the DB are all well-designed. A few items to address before merging:

### 🔴 Critical/Security

1. **Weak key derivation (`encryption.py:36-38`)** — Bare `hashlib.sha256()` provides zero stretching. If `DB_ENCRYPTION_KEY` is ever a passphrase rather than full-entropy random bytes, this is brute-forceable. Use `PBKDF2`/`HKDF`/`scrypt`, or at minimum document and enforce that the key must be generated via `openssl rand -base64 32`.

2. **Thread-unsafe global state (`encryption.py:21-22`)** — `_fernet_instance` and `_encryption_key` are module-level globals mutated without a lock. Concurrent calls to `_get_fernet()` during startup in async/multi-worker environments can race. Wrap in a `threading.Lock`.

3. **`decrypt()` doesn't catch `InvalidToken` (`encryption.py:64`)** — If the Fernet key is rotated or a value is corrupted, `f.decrypt()` raises `cryptography.fernet.InvalidToken` and crashes the request. Catch it, log a warning, and either return the raw value or raise a domain-specific error.

### 🟡 Improvements

4. **Dead fallback code (`routes.py:2484, 2556`)** — `encrypt(user_message) or user_message` — `encrypt()` never returns `None` for non-None input, so the `or` branch is unreachable. Remove for clarity.

5. **Empty string inconsistency (`encryption.py:50` vs `init_db.py:530`)** — New writes encrypt empty strings, but the migration query skips them (`content != ''`). Pick one behavior.

6. **Missing logging in migration** — Email recipients migration block doesn't log its count (all other blocks do). Also, `import json` is inside the loop body at `init_db.py:576`.

7. **pgcrypto enabled but unused** — Consider removing until actually needed, or add a TODO.

### ✅ What's Good

- `enc::` prefix for backwards-compatible gradual migration — no schema changes, no downtime
- HMAC for password reset tokens (deterministic lookup without storing recoverable values)
- Clean separation in `encryption.py`
- Idempotent migration with `NOT LIKE 'enc::%'` guards

nvasilevv:
All review items addressed in 944d69b — looks good, ready to merge. ✅

- HKDF key derivation ✅
- Thread-safe Fernet singleton ✅
- `InvalidToken` graceful handling ✅
- Dead fallback code removed ✅
- Empty string consistency ✅
- Migration logging + `import json` cleanup ✅

PR #6, opened 2026-04-09
Add DocumentViewer enhancements, ChartDesigner, and Knowledge Graph

nvasilevv:
## Summary

- **DocumentViewer**: Enhanced with wikilink support (`![[chart:UUID]]` and `[[note_id:UUID|Title]]`), remark plugins for chart embeds and internal note links, and graph panel integration
- **ChartDesigner**: Full chart design system with JSON config storage (`Chart_Definitions` table), CRUD API, Recharts-based renderer, and Plotly/Kaleido backend for PDF export. Includes chart schema mapper for Recharts↔Plotly conversion
- **Knowledge Graph**: Obsidian-style knowledge graph with `GraphNode`/`GraphEdge` tables, `Note`/`NoteRevision` models (versioned content), wikilink parser service, `react-force-graph-2d` visualization, and graph API endpoints
- **PDF Export**: Extended report renderer with wikilink substitution — chart wikilinks are replaced with Plotly-generated SVGs, note wikilinks rendered as styled text
- **Frontend Pages**: New `/notes` and `/charts` pages with sidebar navigation

### Backend (13 files, ~1,100 lines)
- 4 new DB models: `ChartDefinition`, `Note`, `NoteRevision`, `GraphNode`/`GraphEdge`
- 2 new API routers: `/api/charts/` (CRUD) and `/api/notes/` (CRUD + graph endpoints)
- New services: `chart_schema_mapper.py`, `wikilink_parser.py`
- Enhanced `report_renderer.py` with wikilink-to-image substitution pipeline

### Frontend (14 files, ~1,850 lines)
- New components: `ChartRenderer`, `KnowledgeGraph`, `GraphPanel`
- New remark plugins: `remark-chart-wikilinks`, `remark-wikilinks`
- New pages: `/notes`, `/charts` with full CRUD UI
- Extended API client with chart/note/graph endpoints
- Added `react-force-graph-2d` dependency

## Test plan
- [ ] Verify `/notes` page loads and supports create/edit/delete
- [ ] Verify `/charts` page loads and supports chart creation with JSON config
- [ ] Test `![[chart:UUID]]` rendering inside DocumentViewer
- [ ] Test `[[note_id:UUID|Title]]` links render as clickable internal links
- [ ] Verify Knowledge Graph visualization renders nodes and edges
- [ ] Test PDF export with embedded chart wikilinks generates correct SVGs
- [ ] Run DB migration to create new tables


nvasilevv:
## PR Review — Add DocumentViewer enhancements, ChartDesigner, and Knowledge Graph

**Summary:** This PR adds a data-source population endpoint for charts, graph node lifecycle management (create/update/delete sync), a note deletion endpoint, security hardening for cross-org graph edge leakage, and version-pinned note wikilink resolution in the PDF renderer. The scope is large (~1,300 additions) spanning 14 files, touching backend API, DB models, services, and a full frontend charts page.

---

### Review Findings

#### 🔴 Critical / Security

**1. N+1 query in `analyst_estimates` populate — `charts.py:209`**
```python
for est in reversed(estimates):
    year_ests = list((await session.exec(
        select(AnalystYearEstimate).where(...)
    )).all())
```
With `limit` defaulting to 100, this fires up to 100 additional queries per request. Use a joined load or a single `IN` query against all fetched estimate IDs:
```python
all_ids = [e.id for e in estimates]
all_year_ests = (await session.exec(
    select(AnalystYearEstimate).where(AnalystYearEstimate.estimate_id.in_(all_ids))
)).all()
year_map = defaultdict(list)
for y in all_year_ests:
    year_map[y.estimate_id].append(y)
```

**2. `xbrl_facts` query has no org scoping — `charts.py:169-183`**
The `EodPrice`, `XbrlFact`, and `XbrlContext` queries in `populate_chart_data` have no `org_id` filter. If these tables are multi-tenant, any authenticated user can retrieve any org's XBRL data by guessing a concept name. Verify whether these tables are globally shared or org-scoped, and add the filter accordingly.

**3. Missing graph node creation on note create**
`delete_note` carefully removes the graph node, but scanning the diff there's no corresponding `GraphNode` creation in the note *create* path — only `create_chart` does this. If notes created before this PR have no graph nodes, wikilinks pointing at them will silently produce no edges. Either confirm the note create endpoint already handles this (outside the diff) or add it.

---

#### 🟡 Improvement

**4. `data_source` string is lossy — `charts.py:226`**
```python
chart.data_source = f"{payload.source_type}:{payload.ticker or payload.issuer_id or payload.concept}"
```
`issuer_id` is a UUID; if `ticker` is `None`, the string becomes `analyst_estimates:<uuid>`. Later code parsing this field (e.g. for re-population) would need to distinguish UUID from ticker. Consider storing as a structured value or at minimum document the format.

**5. Graph node label not updated on note title change — `notes.py`**
`update_chart` syncs the `GraphNode.label` when the title changes, but there's no equivalent in the note update endpoint. The knowledge graph will show stale note titles after a rename.

**6. Frontend `load()` has no error state — `charts/page.tsx:575-582`**
```tsx
const load = useCallback(async () => {
  try {
    const data = await getCharts();
    setCharts(data);
  } finally {
    setLoading(false);
  }
}, []);
```
If the API call rejects, `charts` stays empty with no visible feedback. Add a caught error state or at minimum a toast.

**7. `save()` doesn't surface API errors — `charts/page.tsx:624-651`**
`updateChart`/`createChart` returning falsy silently drops the error. The user has no indication the save failed.

**8. `__tablename__ = "note"` addition — `note.py:342`**
This is safe only if SQLModel was already resolving the table name to `"note"`. If it was previously using a different convention (e.g. `"notes"`), this is a silent schema change. Confirm the existing migration target matches.

---

#### 🔵 Nitpick

**9.** `_ALLOWED_RENDER_FORMATS` whitelist is a good pattern, but the error message format is inconsistent with other 422s in the codebase.

**10.** `GraphEdge` imported at module level in `charts.py:17` but only used in delete/update paths. Could be a lazy import, but not a blocker.

**11.** `window.confirm()` for delete — `charts/page.tsx:654`. Works, but other parts of the codebase may use a modal dialog pattern. Consider aligning.

---

### Positive Feedback

- The **org-scoped edge filtering** fix in `get_graph` (`visible_edges`) and the `org_filter` in `_sync_edges` are exactly the right defense-in-depth approach — these close a real cross-tenant data-leakage vector.
- **`output_format` rename** from `format` (which shadows a Python builtin) is a clean, correct fix.
- The **version-pinned wikilink** logic in `_substitute_note_wikilinks` with the dual-key lookup (`uuid@rev_N` falling back to `uuid`) is well-thought-out and preserves historical report accuracy.
- **`await session.flush()`** before creating the graph node in `create_chart` correctly ensures the chart `id` is populated before being referenced as `ref_id`. Good SQLAlchemy discipline.

---

### Conclusion

**Request Changes.** The N+1 query on `analyst_estimates` is a correctness/performance blocker at any realistic `limit`, and the missing org-scoping on XBRL/EOD queries is a potential tenant isolation issue that needs confirmation. The missing graph-node sync on note title updates is a functional gap. Once those three are addressed, this is mergeable.

nvasilevv:
## Addressed in 8555ffe

### Critical fixes

**1. N+1 query on analyst_estimates** — Replaced the per-estimate loop with a single batch `IN` query via `AnalystYearEstimate.estimate_id.in_(est_ids)`, grouped into a `defaultdict(list)` map. Now 2 queries total regardless of limit.

**2. Org scoping on EOD/XBRL queries** — Confirmed: `EodPrice` and `XbrlFact`/`XbrlContext` are globally shared reference tables with no `org_id` column — they contain public market data keyed by ticker/filing. Added a docstring to `populate_chart_data` clarifying this. `AnalystEstimate` is scoped by `issuer_id` which is already access-controlled upstream.

**3. Missing graph node on note create** — This is already handled in the original diff at `notes.py:187-193` (the `create_note` endpoint creates a `GraphNode(node_type="note")` before calling `_sync_edges`). It was part of commit `5934709`, not the latest commit, so it may not have been visible in the incremental diff.

### Improvements

**4. `data_source` string is lossy** — Now also stores structured `_source_meta` dict in `config_json` with explicit `type`, `ticker`, `issuer_id`, `concept`, and `limit` fields for re-population.

**5. Graph node label on note title change** — Already handled in both code paths of `update_note`: lines 302-304 (when content changes) and lines 315-324 (title-only update). Same situation as point 3.

**6 & 7. Frontend error states** — Added `error` state to `ChartsPage`. `load()` now catches failures and shows an error banner. `save()` surfaces API failures via `setDataError()` instead of silently dropping.

**8-11. Nitpicks** — Acknowledged. These are minor and can be addressed in a follow-up if needed.

nvasilevv:
All three critical blockers from the previous review are resolved in the latest commit. ✅

- **N+1 query** — Batch `IN` query with `year_map` dict assembly. Clean fix.
- **Org scoping** — Clarified via docstring: `EodPrice`/`XbrlFact` are globally shared public market data tables with no `org_id` column. No tenant isolation issue.
- **`data_source` ambiguity** — Structured `_source_meta` now stored in `config_json` for unambiguous re-population.
- **Frontend error states** — Both `load()` and `save()` surface errors to the user.

Also noted: graph node creation on note create and label sync on note update were already present in the original diff — my earlier comment was incorrect on those two.

Remaining items are nitpicks (non-blocking):
- Confirm `__tablename__ = "note"` matches the existing migration target.
- `window.confirm()` for delete — minor UX, not a blocker.

**Approve.** Ready to merge.

PR #2, opened 2026-04-09
Add user-visible token budget indicator

nvasilevv:
## Summary

- **Backend**: Expose `token_budget`, `tokens_used`, and `tokens_remaining` on the `GET /api/profile` and `PATCH /api/profile` endpoints so regular users can see their own usage
- **Frontend**: New `TokenBudgetIndicator` component in the sidebar (above user menu) showing a color-coded progress bar with remaining/total tokens
- Hidden for users with unlimited budget (budget = 0); collapsed sidebar shows a compact icon with tooltip

## Details

| File | Change |
|------|--------|
| `backend/app/api/profile.py` | Added 3 token fields to `ProfileResponse`, populated in both GET and PATCH handlers |
| `frontend/src/lib/types.ts` | Added `token_budget`, `tokens_used`, `tokens_remaining` to `UserProfile` type |
| `frontend/src/components/layout/token-budget-indicator.tsx` | New component — progress bar with green/amber/red thresholds at 70%/90% |
| `frontend/src/components/layout/sidebar.tsx` | Render `TokenBudgetIndicator` above `UserMenu` |

## Test plan

- [ ] Verify `/api/profile` returns token budget fields for authenticated users
- [ ] Set a token budget via admin panel, confirm the indicator appears in the sidebar
- [ ] Confirm indicator is hidden when budget is 0 (unlimited)
- [ ] Check color transitions: green (<70%), amber (70-90%), red (>90%)
- [ ] Verify collapsed sidebar shows icon-only indicator with tooltip
- [ ] Confirm no regressions on admin token management UI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7, opened 2026-04-10
Add wikilink autocomplete for notes editor

nvasilevv:
## Summary

- **Wikilink autocomplete**: Typing `[[` in the notes editor opens a searchable dropdown listing notes, filings, and charts. Typing `![[` filters to charts only (embed semantics). Selecting an item inserts the correct wikilink syntax (`[[note_id:UUID|Title]]`, `[[filing:UUID|Title]]`, `![[chart:UUID]]`).
- **New `[[filing:UUID|Display]]` wikilink syntax**: Filings/source documents can now be linked from notes, alongside existing note and chart links. Backend parser, remark plugin, and graph edge sync all updated.
- **Unified search endpoint**: `GET /api/notes/search/references?q=...&types=note,chart,filing` searches across notes (by title), charts (by title), and filings (by ticker/company name). Returns a consistent `{ id, title, type, subtitle }` shape for each match.

### Backend
- `backend/app/services/wikilink_parser.py` — Added `FilingLink` dataclass, `_FILING_LINK_RE` regex, and `extract_filing_links()` function
- `backend/app/api/notes.py` — New `/search/references` endpoint; `_sync_edges()` now creates graph edges for `[[filing:...]]` links targeting `node_type="source"` graph nodes

### Frontend
- `frontend/src/components/notes/wikilink-autocomplete.tsx` — **New component**: positioned dropdown with 200ms debounced search, results grouped by type (notes/filings/charts), keyboard navigation (Arrow keys, Enter, Escape), click selection
- `frontend/src/app/[locale]/notes/page.tsx` — Integrated autocomplete into textarea: detects `[[`/`![[` triggers, computes caret position via mirror div, inserts wikilink on selection
- `frontend/src/lib/remark-wikilinks.ts` — Extended to parse `[[filing:UUID|Display]]` alongside note links; both patterns sorted by position before splicing
- `frontend/src/lib/api.ts` — Added `searchReferences()` API function
- `frontend/src/lib/types.ts` — Added `ReferenceItem` type

## Test plan

- [ ] Open a note for editing, type `[[` — dropdown appears with notes, filings, and charts
- [ ] Type a company ticker (e.g. "AAPL") — matching filings appear
- [ ] Type part of a note title — note results narrow
- [ ] Arrow-key navigate and press Enter — correct wikilink syntax inserted at cursor
- [ ] Type `![[` — only charts shown in dropdown
- [ ] Press Escape — dropdown closes, no text modified
- [ ] Save a note containing `[[filing:UUID|...]]` — backend parses it and creates graph edges
- [ ] Verify existing `[[note_id:...]]` and `![[chart:...]]` links still work


nvasilevv:
## PR #7 Review: Add wikilink autocomplete for notes editor

### 1. Summary

This PR adds a wikilink autocomplete system to the notes editor. Typing `[[` opens a searchable dropdown listing notes, filings, and charts, while `![[` filters to charts only. A new unified backend search endpoint and `[[filing:UUID|Display]]` wikilink syntax are introduced, with graph edge sync and remark plugin updates to support filing links.

---

### 2. Review Findings

#### 🔴 Critical/Security

| # | File | Finding |
|---|------|---------|
| 1 | `backend/app/api/notes.py:199` | **SQL injection via `ilike`**. The `query` variable is interpolated directly into `ilike(f"%{query}%")`. While SQLAlchemy parameterizes the value (preventing true SQL injection), **the `%`/`_` wildcard characters are not escaped**, allowing a user to craft queries like `%` that match all records. This applies to lines 199, 217, 237-239. Use `query.replace("%", "\\%").replace("_", "\\_")` or a helper to sanitize LIKE wildcards. |
| 2 | `backend/app/api/notes.py:228-252` | **Missing org/user scoping on filings search**. Notes and charts are scoped to `user_email == email` or `org_id == org_id`, but the filing search only filters on `status == "completed"` — any authenticated user can search and discover filings from any org. If filings are org-scoped, add an org filter. |

#### 🟡 Improvement

| # | File | Finding |
|---|------|---------|
| 3 | `notes/page.tsx:137-139` | **Dead branch in chart wikilink logic**. Both sides of the ternary produce the identical string `![[chart:${item.id}]]`. The original intent was likely to insert `[[chart:...]]` (without `!`) when the trigger is `[[` rather than `![[`, but currently a chart always gets the embed prefix. |
| 4 | `notes/page.tsx:78-123` | **`detectAutocomplete` depends on `autocomplete` state** (included in its `useCallback` deps). Every time `autocomplete` changes, the callback is recreated, which may cause unnecessary re-renders. Consider removing the state read and using a functional pattern or ref instead. |
| 5 | `wikilink-autocomplete.tsx:104-111` | **O(n²) grouping**. The `grouped.find()` inside the loop is O(n) per item. For the current `limit_per_type=7` (max 21 items) this is fine, but a `Map` would be cleaner. Minor. |
| 6 | `wikilink-autocomplete.tsx:62-86` | **Global keydown capture**. The listener uses `capture: true` on `document`, which intercepts Enter/ArrowDown/ArrowUp globally while the autocomplete is open. If the user is focused on another element (unlikely but possible), keypresses get swallowed. Consider scoping to the textarea or checking `document.activeElement`. |
| 7 | `backend/app/api/notes.py:166-172` | **Return type is `list[dict]`** — consider using a Pydantic response model (`list[ReferenceResponse]`) for consistency with the rest of the API and automatic OpenAPI schema generation. |
| 8 | `notes/page.tsx:35-75` | **Mirror div for caret position is created/destroyed on every keystroke**. Consider caching or using a library like `textarea-caret` to avoid DOM thrashing. |

#### 🔵 Nitpick

| # | File | Finding |
|---|------|---------|
| 9 | `remark-wikilinks.ts:140` | `return index + newNodes.length` — the `visit` callback return value to skip newly-inserted nodes is correct but could benefit from a brief inline comment explaining why, as this is a non-obvious `unist-util-visit` pattern. |

---

### 3. Positive Feedback

- **Solid wikilink parser design** (`wikilink_parser.py`) — clean dataclasses, compiled regexes, and graceful handling of malformed UUIDs. Well-structured and easy to extend.
- **Graph edge sync** (`_sync_edges`) is well-thought-out — batches target node lookups, scopes to org to prevent cross-org edges, and does a clean diff (delete stale, insert new). This is the kind of careful incremental sync that avoids full-delete-reinsert pitfalls.
- **Autocomplete UX details** — 200ms debounce, keyboard nav with arrow keys, outside-click-to-close, newline-breaks-trigger detection, and `requestAnimationFrame` for cursor restoration all show attention to user experience.
- **Remark plugin** handles interleaved note and filing links correctly by sorting replacements by position before splicing — prevents the offset drift bug that commonly breaks multi-match text replacements.

---

### 4. Conclusion

**Request Changes** — Primarily for finding #2 (filing search missing org scoping, potential data leak across orgs) and finding #1 (LIKE wildcard injection). Finding #3 (dead branch) should also be addressed as it indicates a logic bug in chart link insertion.

nvasilevv:
## PR Review — Wikilink Autocomplete

### Summary

This PR adds a wikilink autocomplete system to the notes editor. Typing `[[` opens a searchable dropdown for notes, filings, and charts; `![[` filters to chart embeds only. It introduces a new `[[filing:UUID|Display]]` wikilink syntax with full backend parser, remark plugin, and graph edge support, plus a unified `/search/references` endpoint.

### Review of Addressed Comments (commit c15a6c3)

| Issue | Fix | Verdict |
|---|---|---|
| LIKE wildcard injection (`%`, `_` in search input) | Escaped with `\%` and `\_` | ✅ Fixed correctly |
| Missing comment on filing scope (filings are public, not org-scoped) | Added clarifying comment | ✅ Good |
| Dead branch: chart via `[[` inserted `![[chart:...]]` (embed) instead of `[[chart:...\|Title]]` (reference) | Now correctly differentiates `[[` vs `![[` trigger | ✅ Fixed correctly |
| Unnecessary re-renders from `autocomplete` in `detectAutocomplete` deps | Removed from deps, uses functional setState pattern | ✅ Fixed correctly |

### Findings

#### 🔴 Critical/Security

- **LIKE escape may be incomplete** — The `query` variable gets `%` and `_` escaped, but SQLAlchemy's `.ilike()` treats `\%` as a literal `%` **only if you pass `escape="\\"` to the call**. Without the escape clause, the backslash may have no special meaning (depends on DB dialect — PostgreSQL does not honour it by default). Verify this works, or add `.ilike(f"%{query}%", escape="\\")` explicitly.

- **No input length limit on `q` parameter** — A very long search string could cause expensive LIKE scans across 3 tables. Consider adding a `max_length` validator on the `q` Query param (e.g., 200 chars).

#### 🟡 Improvement

- **Mirror div for caret positioning is created/destroyed on every keystroke** — `getCaretCoords()` creates a DOM element, appends it, measures, and removes it on every keypress. Works but could be a perf concern for large notes. Not blocking.

- **`flatItems.find()` inside render loop** — In `wikilink-autocomplete.tsx`, `flatItems.find((f) => f.item === item)` runs for every item during render (O(n²)). With max 21 items it's fine, but a `Map` would be cleaner.

- **Debounce doesn't cancel in-flight fetches** — `fetchResults` is async and not aborted on unmount. If the component unmounts mid-fetch, `setResults`/`setLoading` will be called on an unmounted component. Consider an `AbortController`.

#### 🔵 Nitpick

- `searchReferences` silently swallows errors (`catch { return []; }`). Fine for autocomplete UX, but worth noting for debuggability.
- Minor style inconsistency: note/chart queries use `|` for OR, filing query uses `or_()`.

### Positive Feedback

- The `detectAutocomplete` backwards scan for unclosed `[[` with newline/`]]` guards handles edge cases well.
- The remark plugin refactor to a unified `Replacement` type with position-based sorting scales nicely to more link types.
- Functional `setState` pattern (`prev => prev ? null : prev`) to avoid unnecessary renders is a nice touch.
- `FilingLink` dataclass + parser mirrors existing patterns perfectly.

### Conclusion

**Comment** — The LIKE escape should be verified for PostgreSQL (may need `escape="\\"`). Everything else looks solid.

nvasilevv:
## Follow-up Review — commit bd5f704

Both fixes from the review are correctly addressed:

1. **LIKE escape clause** ✅ — `escape="\\"` is now passed to all 5 `.ilike()` calls (notes title, charts title, issuer ticker, company_name, company_name_en). PostgreSQL will correctly treat `\%` and `\_` as literal characters.

2. **Input length limit** ✅ — `max_length=200` on the `q` Query param. FastAPI/Pydantic will return a 422 validation error for anything longer, preventing expensive LIKE scans.

Both fixes are minimal and targeted — no unnecessary changes. LGTM.

PR #8, opened 2026-04-11
Render wikilinks as clickable links in notes preview

nvasilevv:
## Summary
- Add `remarkWikilinks` and `remarkChartWikilinks` plugins to the notes page preview `ReactMarkdown`, so wikilink syntax (`[[note_id:UUID|Title]]`) is parsed instead of displayed as raw text
- Render `note-link` elements as styled, clickable spans that navigate to the linked note (looks up loaded notes first, falls back to API fetch)
- Render `filing-link` elements as styled spans showing the display text

## Test plan
- [ ] Open a note containing wikilinks and toggle to preview mode — verify links render as styled text, not raw `[[...]]` syntax
- [ ] Click a note link — verify it opens the linked note
- [ ] Switch to edit mode — verify raw wikilink syntax is still visible in the textarea
- [ ] Type `[[` in the editor — verify autocomplete still works as before


nvasilevv (commented):
## PR Review: Render wikilinks as clickable links in notes preview

### Summary

This PR adds wikilink autocomplete and rendering to the notes editor. When typing `[[` or `![[`, a dropdown appears to search and insert links to notes, filings, and charts. In preview mode, wikilink syntax is parsed and rendered as styled, clickable spans that navigate to the linked entity. A new backend endpoint (`GET /api/notes/search/references`) powers the autocomplete search. Also includes an unrelated CSS fix for zoom-out empty space.

---

### Review Findings

#### 🔴 Critical / Security

1. **Filing search has no org-level authorization boundary** — `backend/app/api/notes.py:95-122`
   The comment says "filings are public exchange documents" but this still means any authenticated user can search all filings across all orgs. If that's intentional, fine — but verify this is the desired policy. The note and chart queries correctly scope to `user_email` or `org_id`, while filings don't. If filings should be scoped, this is a data leak.

2. **`types` query parameter is not validated** — `backend/app/api/notes.py:34`
   The `types` parameter accepts arbitrary comma-separated strings. If someone passes `types=admin,internal`, it just silently ignores unknown types, which is fine. But there's no explicit allowlist validation. Consider validating against `{"note", "chart", "filing"}` to be explicit and reject invalid input with a 422.

#### 🟡 Improvement

3. **Mirror div for caret position is created on every keystroke** — `frontend/src/app/[locale]/notes/page.tsx:271-311`
   `getCaretCoords()` creates a DOM element, appends it to `document.body`, measures it, and removes it — every time the user types a character while autocomplete is active. This is expensive DOM thrashing. Consider caching the mirror div or only computing coordinates when autocomplete state changes (not on every keystroke).

4. **`detectAutocomplete` fires 3 times per keystroke** — `notes/page.tsx:406-409`
   It's called via `onChange` (inside `requestAnimationFrame`), `onKeyUp`, and `onClick`. When typing, both `onChange` and `onKeyUp` fire, plus the rAF. This means `getCaretCoords()` (with its mirror div) can run 2-3 times per character. Consider removing the `onKeyUp` handler since `onChange` + rAF already covers typing, and keeping `onClick` for mouse repositioning only.

5. **`searchReferences` swallows all errors silently** — `frontend/src/lib/api.ts:672-674`
   `catch { return []; }` hides network errors, 401s, 500s, etc. A user with an expired token will just see "No results found" with no indication of why. At minimum, log the error to console.

6. **Keyboard event listener captured at document level** — `wikilink-autocomplete.tsx:553`
   Using `{ capture: true }` on document-level keydown means the autocomplete intercepts **all** keyboard events before they reach other handlers while open. If the user presses Enter for any other reason (e.g. there's a dialog), it will be swallowed. The capture phase listener is aggressive — consider using bubble phase or scoping to the textarea.

7. **`flatItems` lookup uses reference equality** — `wikilink-autocomplete.tsx:618`
   `flatItems.find((f) => f.item === item)` uses object reference equality. This works because `results` and `grouped` share the same object references, but it's fragile — if the grouping logic ever creates copies, this breaks silently. Consider using index-based lookup instead.

8. **No debounce cancellation on unmount race** — `wikilink-autocomplete.tsx:510-516`
   `fetchResults` is async. If the component unmounts mid-flight, `setResults` and `setSelectedIndex` will be called on an unmounted component. Consider an abort controller or a mounted ref guard.

9. **Mixed concerns in the PR** — The CSS zoom fix (`website/src/index.css`) and `overflow-x: hidden` changes are unrelated to wikilinks. These should ideally be in a separate commit/PR for cleaner history, though they are already in a separate commit.

#### 🔵 Nitpick

10. **`remarkChartWikilinks` imported but not shown in diff** — `notes/page.tsx:252`
    `import { remarkChartWikilinks } from "@/lib/remark-chart-wikilinks"` — this import is new in the diff but the file itself isn't changed. Verify this module exists and exports the plugin correctly.

11. **`as any` casts on ReactMarkdown components** — `notes/page.tsx:434,457`
    The double `as any` cast (`{ ... } as any`) to wire custom HTML elements into ReactMarkdown's `components` prop is a bit rough. If `react-markdown`'s types don't support custom elements, a `// @ts-expect-error` with an explanation is more honest.

12. **Filing link is not clickable** — `notes/page.tsx:452-456`
    Note links navigate to the linked note on click, but filing links are just styled spans with no click handler. If this is intentional (filings open from the feed, not inline), add a comment explaining why. Otherwise, this is a missing feature.

---

### Positive Feedback

- **LIKE wildcard escaping** is done correctly and consistently across all three entity types, with `escape="\\"` on every `ilike()` call. The `max_length=200` guard on the query param is a nice touch.
- **The remark plugin refactor** is clean — generalizing `Replacement` to hold a `WikilinkNode` union type and sorting replacements by position avoids interleaving bugs when both note and filing links appear in the same text node.
- **Keyboard navigation** in the autocomplete is well-implemented with proper index clamping and `mouseEnter`/`mouseDown` handling.
- **Debounced search** (200ms) is a sensible default that balances responsiveness with API load.
- **Graph edge sync** correctly extends to filing links, maintaining consistency in the knowledge graph.

---

### Conclusion

**Request Changes** — The main concern is item #3/#4 (DOM thrashing on every keystroke) and item #6 (capture-phase keyboard listener). The filing authorization model (#1) needs explicit confirmation that it's intentional. No security blockers, but the performance issues will degrade UX in longer notes.

nvasilevv:
Reviewed commit `9689aad` — all major findings addressed. Here's the scorecard:

| # | Finding | Status |
|---|---------|--------|
| 2 | `types` param not validated | ✅ Intersects with allowlist, returns `[]` if empty |
| 3 | Mirror div recreated every keystroke | ✅ Cached in `mirrorRef`, cleaned up on unmount |
| 4 | `detectAutocomplete` fires 3x per keystroke | ✅ `onKeyUp` removed |
| 5 | Errors swallowed silently | ✅ `console.error` added |
| 6 | Capture-phase keyboard listener | ✅ Switched to bubble phase |
| 7 | Reference equality in flatItems | ✅ `Map<string, number>` keyed by `type-id` |
| 8 | No unmount guard on async fetch | ✅ `mountedRef` guard added |
| 12 | Filing link not clickable | ✅ Comment added explaining intent |

Remaining items are intentional design decisions (#1 filing auth, #9 mixed commits) or nitpicks (#10, #11). LGTM — good to merge.

PR #9, opened 2026-04-12
Add agent tools for research notes with wikilink-aware writing

nvasilevv:
## Summary
- Wires `read_note`, `list_notes`, `search_notes`, `create_note`, and `update_note` into the chat agent so it can browse and author research notes directly.
- `search_notes` matches both title and latest-revision content (ilike, escaped wildcards, top 10 with snippet).
- `create_note`/`update_note` reuse helpers from `app.api.notes` so knowledge-graph edges are rebuilt from wikilinks on every revision.
- System prompt now requires wikilink syntax (`[[note_id:UUID|Title]]`, `![[chart:UUID]]`, `[[filing:UUID|Title]]`) whenever the agent writes note content, and forbids fabricating UUIDs.

## Test plan
- [ ] Ask the agent to list and search notes; confirm tool calls show up and return expected ids/snippets.
- [ ] Ask the agent to create a note that references another note and a chart; verify the saved content uses wikilink syntax and that graph edges appear in `/notes/graph/full`.
- [ ] Ask the agent to update an existing note with new wikilinks; confirm a new revision is created and stale edges are removed.
- [ ] Confirm org-scoping: notes from another user/org are not returned.


nvasilevv:
# PR Review

**Branch:** `feat/agent-research-notes` → `main`
**Files:** `backend/app/api/routes.py` (+256/-5), `backend/app/services/genui_tools.py` (+139)

## Summary
Wires five research-note tools (`list_notes`, `search_notes`, `read_note`, `create_note`, `update_note`) into the chat agent's tool loop, plus a system-prompt block teaching the model the wikilink syntax (`[[note_id:UUID|Title]]`, `![[chart:UUID]]`, `[[filing:UUID|Title]]`) so authored notes feed the knowledge graph. The executors mirror the existing `app/api/notes.py` REST handlers and reuse `_get_note_or_404` / `_sync_edges` / `_serialize_note`.

## Review Findings

### 🔴 Critical / Security
None blocking. Auth is enforced — every code path scopes by `email`/`org_id` from the chat closure and reads go through `_get_note_or_404`. `_sync_edges` already prevents cross-org edge creation. The substring search escapes `%` and `_` and uses `escape="\\"` properly, so no LIKE-injection. No new dependencies, no PII logging beyond title (consistent with existing routes).

### 🟡 Improvement
- **Significant DRY violation (`routes.py:2445-2674`).** `execute_list_notes`, `execute_read_note`, `execute_create_note`, `execute_update_note` are line-for-line copies of the same logic in `backend/app/api/notes.py:312-436`, including the correlated-subquery latest-revision pattern and the edge-rebuild dance. Two copies of this will drift — e.g. when someone fixes a bug in `_sync_edges` callers or tweaks the visibility filter, only one side will get it. **Recommended fix:** extract `notes.py`'s body into service functions (e.g. `app/services/notes_service.py` with `list_notes_for_user`, `search_notes_for_user`, `create_note`, `update_note`) and have *both* the REST routes and the agent executors call them. The PR is otherwise straightforward but this is the meat of the maintainability concern.
- **`search_notes` content-only matches require a NoteRevision row.** The OUTER join + `or_(title.ilike, content.ilike)` works for title hits even when no revision exists, but content hits depend on the join. Fine; just note the snippet builder at `routes.py:2528-2536` will produce `content[:200]` (empty) if `rev` is None and the user had a title-only match — minor cosmetic issue.
- **Visibility filter quirk inherited from `notes.py:334`.** `((Note.org_id == org_id) if org_id else False)` evaluates to a literal Python `False` when `org_id is None`, which SQLAlchemy translates to `false` in SQL — semantically fine, but the same pattern exists at `routes.py:2475` and `routes.py:2509`. If you do the service-extraction above, this gets fixed in one place.
- **No length caps on agent-supplied `title` / `content`.** The model can write arbitrarily large notes into the DB. The REST API has the same gap, so not a regression, but worth flagging while you're in here — a 100KB title is plausible model misbehavior.
- **`note_type` is a free-form string from the model.** `execute_create_note` defaults to `"research"` but accepts anything. If you have a known set of note types, validate against an enum to prevent typos like `"reasearch"` polluting your taxonomy.

### 🔵 Nitpick
- **Imports inside function bodies** (`from uuid import UUID as _UUID`, `from app.api.notes import …`, etc., at `routes.py:2446-2448, 2467-2468, …`). `routes.py` already imports plenty at module top — move these up unless you specifically need lazy import to break a cycle.
- **System-prompt bloat (`routes.py:1129-1144`).** ~16 new lines added to an already very long prompt. Consider whether the wikilink rules can live solely in the per-tool `description` strings of `CREATE_NOTE_TOOL` / `UPDATE_NOTE_TOOL` (you already have `_WIKILINK_GUIDANCE` there). Tool descriptions are visible to the model when the tool is invoked and keep the system prompt slimmer.
- **`logger.info` includes `title=%r`** (`routes.py:2519, 2575`). Probably fine, but research-note titles can be sensitive — confirm against your logging policy.
- **`update_note` graph-node label sync is split** (`routes.py:2625-2638` for the content-changed branch and `2640-2654` for the title-only branch). The two paths duplicate the `select(GraphNode)` lookup — combine.
- **`# type: ignore[union-attr]`** sprinkled through the new queries — same as the existing file, so not new noise, just inherited.

## Positive Feedback
- Reusing `_sync_edges` means notes the agent writes immediately become first-class graph citizens — good architectural alignment.
- `_WIKILINK_GUIDANCE` constant in `genui_tools.py:597` keeps the wikilink contract DRY between the two write tools' descriptions.
- The "search notes by substring before linking, never fabricate UUIDs" instruction in the system prompt is exactly the right guardrail against the LLM hallucinating IDs into the knowledge graph.
- The substring escaping in `execute_search_notes` (`routes.py:2497`) is a nice touch — easy thing to forget.
- `search_notes` snippet windowing (`routes.py:2528-2536`) gives the model useful context without dumping full revisions into the tool result.

## Conclusion
**Request Changes** — non-blocking but the duplication of the entire notes-CRUD logic into `routes.py` is the kind of thing that will silently rot. Extract a shared `notes_service` module, point both the REST handlers and the new agent executors at it, then merge. Everything else (length caps, `note_type` validation, in-function imports, prompt slimming) is improvement-tier and can land in this PR or a follow-up at your discretion.

nvasilevv:
# Follow-up Review (commit d41ffcf)

## Summary
The "Extract notes_service" commit nails the main concern from the previous review: ~250 lines of duplicated CRUD logic now live in a single `app/services/notes_service.py`, and both `app/api/notes.py` and the agent executors in `app/api/routes.py` are thin callers. Net diff is +430 / -447, so the refactor actually came out shorter than what it replaced. 

## What got addressed

- 🟡 **DRY violation** — Fixed. `notes_service.py` owns `get_note`, `get_latest_revision`, `get_revision`, `list_notes`, `search_notes`, `create_note`, `update_note`, `delete_note`, and `sync_edges`. Both call sites are now ~5-line wrappers. The REST `delete_note` route was also pulled into the service in passing — nice opportunistic cleanup.
- 🟡 **Visibility filter quirk** — Fixed. `_visibility_clause` (`notes_service.py:39`) does a clean if/else: `or_(user_email==email, org_id==org_id)` when `org_id` is set, else `user_email==email`. The literal-`False` SQL artifact is gone.
- 🟡 **search_notes snippet builder** — Fixed (`routes.py:2487-2500`). Title-only matches now fall back to the note title instead of returning an empty string.
- 🔵 **Combined GraphNode lookup in update_note** — Fixed. `notes_service.update_note` (`notes_service.py:182-211`) does a single `select(GraphNode)` that handles both the title-only and content-changed branches.
- 🔵 **In-function imports** — Fixed. `from app.services import notes_service` and `from app.services.notes_service import NoteNotFoundError, serialize_note as _serialize_note` are now at the top of `routes.py`. The lazy `from uuid import UUID as _UUID` and `from app.models.note_revision import NoteRevision` blocks inside each executor are gone.
- 🔵 **Title in logging** — Fixed. `create_note` and `search_notes` log `len(content)` / `query_len` instead of the raw title/query string.
- ➕ **Bonus: domain-specific exception.** Replacing `HTTPException`-bubbling-through-services with `NoteNotFoundError` is the right call — services shouldn't know about HTTP. The REST handlers now catch and re-raise as 404, and the agent executors catch and return `{"error": ...}`. Clean separation.

## Still outstanding (acknowledged as deferred, but worth tracking)

- 🟡 **No length caps on `title` / `content`.** Not a regression and matches the existing REST behavior, but the agent introduces a new way to write arbitrarily large notes (a misbehaving model can spam content). Worth a follow-up — even a 64KB content cap and 256-char title cap would prevent the worst tail.
- 🟡 **`note_type` is still a free-form string.** `notes_service.create_note` accepts whatever the model passes. If there's a known set of valid types, an enum check here would be cheap insurance against typos polluting the taxonomy.
- 🔵 **System-prompt wikilink block (`routes.py:1129-1144`)** still adds ~16 lines to the system prompt even though `_WIKILINK_GUIDANCE` is already in the per-tool descriptions. Not blocking, but consider trimming on the next pass.

## Nitpicks on the new code

- 🔵 **`_parse_note_uuid` return-type union** (`routes.py:2447-2451`): returns `UUID | dict[str, Any]` and the caller does `if isinstance(nid, dict): return nid`. It works, but the type union is awkward. A small `InvalidNoteIdError` (mirroring `NoteNotFoundError`) and a `try/except` at each call site would read more cleanly and match the pattern you just established for not-found.
- 🔵 **`notes_service.sync_edges`** still does `from app.services.wikilink_parser import …` inside the function. Not introduced by this PR (it was already lazy in the original `_sync_edges`), but now that everything else moved to top-of-module imports, this is the lone stray. Worth checking whether the lazy import is actually breaking a circular dependency — if not, hoist it.
- 🔵 **`notes_service.create_note` no longer defaults `title` / `content` to `""`.** The agent executor still does `title or ""` / `content or ""` before calling, but the REST `create_note` route relies on the Pydantic model to enforce non-null. Slightly inconsistent surface; not a bug, just something to be aware of if a third caller ever shows up.

## Conclusion
**Approve.** The refactor cleanly addressed every blocking item from the previous review and several of the nitpicks too. The deferred items (length caps, `note_type` enum, prompt slimming) are fine to land in a follow-up. Ship it.

PR #5, opened 2026-04-09
Add Notes and Charts pages with sidebar navigation

nvasilevv:
## Summary
- **Notes page** (`/notes`) — list + markdown editor with live preview, create/edit/save with auto-versioning, Knowledge Graph trigger button wired in
- **Charts page** (`/charts`) — list + chart designer with JSON data editor, live Recharts preview, chart type/axis config, delete, embed code display
- **Navigation** — sidebar entries for Notes (StickyNote) and Charts (BarChart3) with EN/BG i18n
- **API layer** — frontend CRUD functions for charts (`getCharts`, `createChart`, `updateChart`, `deleteChart`) and notes (`getNotes`, `getNote`, `createNote`, `updateNote`, `getNoteRevisions`)
- **Types** — `ChartDefinition`, `NoteItem`, `NoteRevisionItem`

These features were built in PR #4 (backend + components) but never exposed to users. This PR wires them into the UI.

## Test plan
- [ ] Navigate to `/notes` — empty state shown, "New Note" button works
- [ ] Create a note with markdown content — saves and shows in list
- [ ] Edit an existing note — revision number increments
- [ ] Toggle preview mode in editor — renders markdown
- [ ] Click Knowledge Graph button in notes header — graph panel opens
- [ ] Navigate to `/charts` — empty state shown, "New Chart" button works
- [ ] Create a chart with sample data — live preview renders correctly
- [ ] Switch chart types (area/bar/composed/line) — preview updates
- [ ] Delete a chart — removed from list
- [ ] Verify sidebar shows Notes and Charts entries in both EN and BG

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4, opened 2026-04-09
Add DocumentViewer, ChartDesigner, and Knowledge Graph

nvasilevv:
## Summary
- **Database**: 5 new models — `Note`, `NoteRevision`, `ChartDefinition`, `GraphNode`, `GraphEdge` — for versioned research notes, JSON-configured charts, and an Obsidian-like knowledge graph
- **ChartDesigner**: Charts stored as JSON configs, rendered by Recharts on the frontend and Plotly+Kaleido for PDF export. Custom `![[chart:UUID]]` wikilink syntax embedded in markdown
- **Knowledge Graph**: `[[note_id:UUID|Title]]` wikilinks with automatic edge extraction on save, `react-force-graph-2d` visualization in a slide-out panel, and version pinning (`@rev_N`) for compliance
- **DocumentViewer**: Table overflow fix for wide financial tables, remark plugins for chart embeds and note links
- **PDF Pipeline**: Wikilink substitution (charts → SVG, note links → bold text) before WeasyPrint rendering

## New API Endpoints
| Method | Path | Purpose |
|--------|------|---------|
| POST/GET/PATCH/DELETE | `/api/charts[/:id]` | Chart definition CRUD |
| GET | `/api/charts/:id/render` | Server-side Plotly SVG/PNG |
| POST/GET/PUT | `/api/notes[/:id]` | Notes CRUD with auto-versioning |
| GET | `/api/notes/:id/revisions` | Revision history |
| GET | `/api/notes/graph/full` | Full knowledge graph (nodes + edges) |
| GET | `/api/notes/graph/neighbors/:id` | 1-hop neighborhood |

## Test plan
- [ ] Start app, verify 5 new tables created in PostgreSQL
- [ ] POST a chart definition, verify JSON roundtrips via GET
- [ ] GET `/api/charts/:id/render` returns valid SVG
- [ ] Create a note with `![[chart:UUID]]` → verify Recharts renders inline
- [ ] Create two notes with `[[note_id:X|Title]]` cross-links → verify `graph_edge` rows created
- [ ] Open graph panel → nodes appear with correct colors (blue/green/yellow)
- [ ] Open a document with wide financial tables → horizontal scroll works
- [ ] Generate PDF with chart wikilink → SVG appears in output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11, opened 2026-04-13
Translate agent [cite-N] tokens to filing wikilinks on note save

nvasilevv:
## Summary

Agent-written research notes used to contain chat-style `[cite-N]` tokens that had no rendering infrastructure on the notes side — they appeared as literal broken text and filings referenced from search_filings never became graph nodes. This PR makes note citations first-class:

- `[cite-N]` tokens are rewritten into `[[filing:UUID#cK|Label]]` wikilinks inside `execute_create_note` / `execute_update_note`, using the in-scope `citation_map` built by `search_filings`.
- A new per-revision `citations_json` sidecar stores the rich locator (filing_id, filename, ticker, filing_date, page_start, section_title, highlight_snippet, label) keyed by the opaque local handle (`c1`, `c2`, ...).
- The notes page resolves the local key from the sidecar and opens `DocumentViewer` with `highlightSnippet`, matching chat's click-to-view behavior. Because the locator is snippet-based, this works for both paginated PDFs *and* markdown filings — no page-semantics assumption.
- Because the rewritten wikilinks still flow through `extract_filing_links`, filings automatically become graph nodes with edges from the note. No graph code changes.

## Changes

**1. Wikilink parser extended** (`backend/app/services/wikilink_parser.py`, `frontend/src/lib/remark-wikilinks.ts`)
- Filing wikilinks now accept an optional opaque `#key` suffix: `[[filing:UUID#c1|Label]]`.
- `FilingLink` dataclass carries `citation_key`; edge extraction ignores it (still one graph edge per filing).
- Frontend `remark-wikilinks` passes `citationKey` through `hProperties`.

**2. `citations_json` sidecar column on `NoteRevision`** (`backend/app/models/note_revision.py`, `backend/app/db/init_db.py`, `backend/app/services/notes_service.py`)
- Nullable JSONB column, added via the existing `DO $$ ... $$` migration block.
- `create_note` / `update_note` service functions accept `citations` and persist per revision.
- `update_note` preserves the prior sidecar if the REST caller supplies content without citations, so manual edits don't strip metadata.
- API serializer exposes `citations` on note responses.

**3. Agent-side cite-to-wikilink translator** (`backend/app/api/routes.py`)
- `_rewrite_citations_to_wikilinks` rewrites `[cite-N]` tokens into `[[filing:UUID#cK|Label]]` wikilinks and builds the per-note sidecar, deduping repeated cites to the same local key.
- Wired into both `execute_create_note` and `execute_update_note`.
- Updated the NOTE WRITING RULES system prompt so the model uses `[cite-N]` consistently across chat and notes and does not hand-author filing wikilinks for search_filings results.
- Tokens without a resolvable `document_id` are left untouched (graceful fallback).
- `read_note` automatically returns the sidecar (via `serialize_note`), so the model can inspect it on updates.

**4. Frontend `filing-link` → `DocumentViewer` wiring** (`frontend/src/app/[locale]/notes/page.tsx`, `frontend/src/lib/types.ts`)
- `NoteItem` gained a `citations` map typed by `NoteCitation`.
- The placeholder `filing-link` renderer is now clickable: resolves the citation key against `selected.citations`, then opens `DocumentViewer` with `documentId`, `filename`, `ticker`, and `highlightSnippet`.
- Viewer mounts as a slide-in panel on the right, mirroring `ask-panel.tsx`.

## Test plan

- [ ] Start a chat, ask a question that triggers `search_filings`, and have the agent `create_note` citing the result. Confirm the saved note's content contains `[[filing:...#c1|...]]` wikilinks (not raw `[cite-1]`).
- [ ] Open the new note in the notes UI. Confirm the filing link is clickable and the DocumentViewer opens with the snippet highlighted.
- [ ] Verify the graph view shows the filing as a node with an edge from the note.
- [ ] Have the agent `update_note` the same note with additional citations and confirm a new revision is written and the sidecar is rebuilt.
- [ ] Edit a citation-bearing note via the REST `PUT /notes/{id}` (manual markdown edit, no citations in payload) and confirm the existing sidecar is preserved.
- [ ] Run the DB migration against a real Postgres and confirm the `note_revision.citations_json` column is added.


nvasilevv:
# PR Review — Translate agent `[cite-N]` tokens to filing wikilinks on note save

## Summary
Wires agent-written `[cite-N]` tokens through a save-time rewriter that produces `[[filing:UUID#cK|Label]]` wikilinks plus a per-revision `citations_json` sidecar, then resolves that sidecar in the notes UI to open `DocumentViewer` with the right snippet. Cleanly reuses the existing wikilink graph extractor — no graph-side changes — and gracefully degrades when metadata is missing. Overall a focused, well-scoped change with one real correctness concern around sidecar preservation on agent updates.

## Findings

### 🔴 Critical / Security
None. `citation_map` values originate from server-side `search_filings` results, not user input, so the rewritten wikilinks aren't an injection vector. `highlight_snippet` ultimately flows into `DocumentViewer` — assumed safe given existing chat usage of the same field, but worth confirming the viewer escapes it (not changed in this PR).

### 🟡 Improvement

1. **Stale sidecar inheritance on agent `update_note`** — `routes.py:2598-2603` + `notes_service.py:224-228`. When the agent calls `update_note` with content that contains *no* `[cite-N]` tokens, `_rewrite_citations_to_wikilinks` returns `citations = {}`, then `citations or None` collapses to `None`. The service then *preserves the previous revision's sidecar*. That's the desired behavior for REST PUTs but wrong for agent updates: the agent intentionally rewrote the note without cites, and the new revision will carry orphaned `c1/c2` keys that no longer match any wikilink in `content`. Suggest distinguishing "caller didn't pass citations" (None) from "caller passed an empty dict" (`{}`) — the agent path should always pass `citations` (even `{}`) so preservation only fires for the REST path.

2. **`import re as _re` inside the closure** — `routes.py:2566`. `re` is already imported at module top (used heavily elsewhere in this file). Move the alias out or just use `re.sub` directly; the local re-import runs on every call and adds noise.

3. **Citation key regex is permissive** — `wikilink_parser.py:24` and `remark-wikilinks.ts:17` accept `[a-z0-9-]+`, but the producer only ever emits `c\d+`. Tightening to `c\d+` (or at least `[a-z]\w*`) would prevent a malformed `[[filing:UUID#-|x]]` from parsing as a "valid" citation key, and makes the contract between producer/consumer explicit.

4. **`_label` falls back to `"Source"` for unlabeled cites** — `routes.py:2538`. Fine, but the label is baked into the wikilink at save time, so a later metadata fix never propagates. Consider deferring label rendering to the frontend (look it up from the sidecar) so labels stay live; the wikilink only needs the citation key. Lower-priority refactor.

5. **`extract_filing_links` now returns `FilingLink` objects with a new field** — `wikilink_parser.py:88-89`. Worth a quick grep to confirm no caller was unpacking `FilingLink` positionally or comparing for equality. Likely fine since it's a `@dataclass` with a defaulted field, but a one-line check would harden the change.

6. **Frontend viewer panel layout** — `notes/page.tsx:513-522`. The slide-in viewer is appended as a flex sibling, which will compress the notes column on narrow viewports rather than overlaying like `ask-panel`. The PR description says "mirroring `ask-panel.tsx`" — worth verifying it actually behaves as an overlay, not an inline column, on `md` widths.

### 🔵 Nitpick

- **Type annotations on `rewritten`/`citations` in `execute_update_note`** (`routes.py:2597-2602`) — the explicit `rewritten: str | None` declaration before the if/else is redundant; assignment in both branches gives mypy enough.
- **`citations or None` idiom** appears 3× — a tiny helper or a clearer comment would help, since the empty-dict-vs-None distinction is load-bearing (see 🟡 #1).
- **Migration block in `init_db.py`** mixes "user token columns" and "note_revision citations_json" inside one `DO $$` block. Functionally fine, just slightly odd grouping.
- **`NoteCitation.filing_id` is required in TS** but the Python sidecar writes `meta.get("document_id")` which is checked truthy first — consistent, but the TS type could mark `page_start` as just `number | null` (not also `undefined`) to match the JSON shape exactly.

## Positive Feedback

- Clean separation of concerns: the rewriter is a pure function over `(content, citation_map)`, the service layer just persists a blob, and the frontend treats the sidecar as opaque locator metadata. Easy to test and reason about.
- Reusing `extract_filing_links` so filings still become graph nodes "for free" is exactly the right call — no parallel graph plumbing to maintain.
- Snippet-based highlighting (vs. page numbers) is the right abstraction for working across PDFs and markdown filings; nice forward-thinking choice.
- The system-prompt update is explicit about *not* hand-authoring filing wikilinks for `search_filings` results, which avoids the exact ambiguity that would have made this feature flaky.
- Sidecar preservation on REST PUT is a thoughtful detail — it's the right default for that path (just needs to be scoped to it; see 🟡 #1).

## Conclusion

**Request Changes** — primarily for 🟡 #1 (the stale sidecar inheritance is a real correctness bug for the agent update path). Everything else is polish. Once the agent path always passes an explicit `citations` value, this is a clean merge.

nvasilevv:
Addressed in 9bd529c:

- **🟡 #1 Stale sidecar on agent update** — fixed. Agent executors now pass `citations` through as-is (even `{}`) rather than collapsing empty dicts to `None`. The service's `None`-vs-`{}` gate correctly distinguishes "agent intentionally rewrote with no cites" (empty dict → no inheritance, stored as NULL) from "REST PUT didn't touch citations" (None → preserve prior). Inheritance now only fires on the REST path as intended.
- **🟡 #2 `import re as _re`** — dropped the alias and hoisted a plain `import re` to the top of the translator helper.
- **🟡 #3 Permissive citation key regex** — tightened to `c\d+` for the *captured* key, but the suffix is still consumed when present so unknown shapes (e.g. `[[filing:UUID#xyz|...]]`) still parse as a valid filing link with `citation_key=None` and still emit a graph edge. Verified with a smoke test across all four shapes (cN, no key, unknown key, key without display text).
- **🟡 #5 `FilingLink` callers** — grepped; the dataclass is only referenced inside `wikilink_parser.py`. Adding a defaulted field is safe.

Leaving these as out-of-scope polish for a follow-up:
- 🟡 #4 (defer label rendering to the frontend) — reasonable refactor, but the current eager-bake matches how chat citations work and keeps the frontend purely consuming the sidecar. Not load-bearing.
- 🟡 #6 (viewer column vs. overlay) — the layout is deliberately the same pattern as `ask-panel.tsx` (flex sibling with `md:w-[480px] lg:w-[560px] shrink-0`). If `ask-panel` behaves as an overlay on `md` widths, so will this; if not, both need the same fix.
- 🔵 nitpicks — left as-is.

nvasilevv:
Re-reviewed `9bd529c` — all three high-priority items are correctly addressed.

**🟡 #1 (the real bug):** Fixed end-to-end. `execute_create_note` / `execute_update_note` now pass `citations=citations` (not `citations or None`), so the empty dict from the rewriter survives. The service layer's `effective_citations is None` gate still triggers inheritance only on the REST PUT path. Traced it: agent → rewriter returns `{}` → service sees `{}` (not None) → no prior-revision lookup → `{} or None` stores NULL. The "agent rewrote without cites" case no longer carries orphan `c1/c2` keys. 👍

**🟡 #2:** `import re` hoisted to the top of `_rewrite_citations_to_wikilinks`, alias dropped, `_replace`'s type hint switched from the string forward-ref to a real `re.Match[str]`. Cleaner than I suggested.

**🟡 #3:** Nice solution — `(?:#(?:(c\d+)|[a-z0-9-]+))?` consumes any suffix shape but only *captures* `c\d+`. So `[[filing:UUID#xyz|Label]]` still parses as a valid filing link with `citation_key=None` and still emits a graph edge, while malformed keys can't masquerade as real ones. Backend and frontend regexes stay in sync.

**🟡 #5:** Acknowledged — `FilingLink` is only referenced inside `wikilink_parser.py`, so the defaulted field is safe.

**Deferred (#4, #6, nitpicks):** Reasonable. The pushback on #6 is fair — if `ask-panel.tsx` uses the same flex-sibling pattern, the layout is consistent with the established convention rather than wrong; better to fix both together if it does need fixing.

Flipping my verdict: **Approve** ✅

PR #12, opened 2026-04-13
Add document IDs as first-class citation targets in research notes

nvasilevv:
## Summary

Research-note citations now support `[[document:UUID#cN|Label]]` wikilinks alongside the existing `[[filing:...]]` shape, and both resolve correctly on the note-rendering side and the knowledge-graph side. Fixes the bug where every citation in agent-written notes was silently dropped from the graph because the cite translator was wrapping document UUIDs in `[[filing:...]]` and `sync_edges` couldn't resolve them.

## Backend

- `backend/app/services/cite_translator.py` (new) — hoisted `rewrite_citations_to_wikilinks` out of `routes.py` closure. Now emits `[[document:UUID#cN|Label]]` and writes both `document_id` and legacy `filing_id` into the sidecar.
- `backend/app/services/wikilink_parser.py` — added `_DOCUMENT_LINK_RE`, `DocumentLink`, `extract_document_links`.
- `backend/app/services/notes_service.py` — `sync_edges` now extracts document links, has a legacy fallback that retries unmatched `[[filing:UUID]]` lookups against the document table, and uses a `public_or_org` filter so org-scoped notes can match public filing/document nodes.
- `backend/app/services/repository.py` — added `_format_source_label`, `_ensure_source_graph_node`, `_ensure_document_graph_node` helpers; seed source nodes in `upsert_filing` and document nodes in `create_document` with shared `"{ticker} {doc_type} {filing_date}"` labels.
- `backend/app/api/routes.py` — closure now delegates to the hoisted module function.
- `backend/app/models/graph.py` — docstring + node_type comment lists `note | source | document | chart`.
- `backend/scripts/backfill_graph_nodes_for_sources.py` (new) — idempotent, `--dry-run` aware: seeds source + document nodes for all existing filings/documents, then re-runs `sync_edges` for every note. Legacy fallback means no markdown is rewritten.

## Frontend

- `frontend/src/lib/remark-wikilinks.ts` — `DOCUMENT_LINK_RE`, `DocumentLinkNode`, `makeDocumentLinkNode`, third pass in the replacements loop.
- `frontend/src/app/[locale]/notes/page.tsx` — new `document-link` renderer; existing `filing-link` renderer also reads `meta.document_id ?? meta.filing_id` for forward-compat.
- `frontend/src/lib/types.ts` — `NoteCitation.document_id` is now the canonical key, `filing_id` kept as legacy alias.
- `frontend/src/components/graph/knowledge-graph.tsx` + `graph-panel.tsx` — node_type union widened to include `"document"`, cyan `#22d3ee` color added.

## Tests

- `backend/tests/test_wikilink_parser.py` — 7 tests (note links, filing+citation key, unknown-suffix tolerance, document parsing, malformed UUID skip, chart embed, parser independence).
- `backend/tests/test_cite_translator.py` — 7 tests (passthrough, single cite, repeat reuse, multiple distinct, unresolvable, missing document_id, label fallback). **No \`[[filing:\` is ever emitted** — explicitly asserted.

## Deploy steps

1. Merge PR.
2. SSH to the server and run inside the backend container:
   \`python -m scripts.backfill_graph_nodes_for_sources --dry-run\` then without \`--dry-run\` once the dry output looks right.
3. Verify the Games Operators note's graph shows edges.

## Two non-obvious things worth flagging for review

- **Filing graph_nodes never existed before** — discovered during investigation. The PR seeds them too, so any pre-existing `[[filing:<real-filing-uuid>]]` wikilink (from autocomplete) starts working as well.
- **Org filter loosened**: org-scoped notes can now reach public (`org_id IS NULL`) filing/document nodes. This is what makes citations work for org users at all and is intentional.

## Test plan

- [x] `test_wikilink_parser.py` — 7/7 passing
- [x] `test_cite_translator.py` — 7/7 passing
- [ ] Run backfill script in `--dry-run` mode on staging, confirm counts look sane
- [ ] Run backfill for real, then open the Games Operators note and confirm graph edges appear
- [ ] Manually create a new agent-written research note, confirm citations render as document links and open `DocumentViewer`
- [ ] Open the knowledge-graph panel, confirm cyan document nodes appear with correct labels


nvasilevv:
# PR Review — "Add document IDs as first-class citation targets in research notes"

## Summary
Hoists the `[cite-N]` → wikilink translator into its own module, introduces `[[document:UUID#cN]]` as a new first-class citation shape alongside `[[filing:...]]`, and wires document-type `graph_node` rows into the knowledge graph (backend + frontend + backfill). Fixes a real bug where agent-written citations were silently dropped from the graph because the translator wrapped **document** UUIDs in `[[filing:...]]` wikilinks that `sync_edges` couldn't resolve. Well-scoped, well-tested (14 new unit tests), with a sound legacy fallback strategy.

## Review Findings

### 🔴 Critical / Security
None. SQL is parameterized throughout, no secrets or PII introduced, no new external deps, authz model (`public_or_org_filter`) is intentional and documented.

### 🟡 Improvement

- **Race condition in `_ensure_source_graph_node` / `_ensure_document_graph_node`** (`backend/app/services/repository.py:324,354`). Both do SELECT-then-INSERT without a unique constraint on `(node_type, ref_id)` — `graph.py:20` shows `GraphEdge` has one but `GraphNode` does not. Two concurrent `create_document` calls for the same doc can both miss the existing row and insert duplicate graph_nodes. Consider adding `UniqueConstraint("node_type", "ref_id", name="uq_graph_node_ref")` to `GraphNode.__table_args__` (plus an Alembic migration), or wrapping the ensure helpers in an `ON CONFLICT DO NOTHING` upsert. Low probability in practice, but cheap to prevent.

- **`upsert_filing` only seeds the graph node on the insert path** (`repository.py:441-455`). If a pre-PR filing somehow lacks a graph node and gets re-upserted, the existing branch at `repository.py:399-417` silently returns without repairing it. The backfill script covers the one-time migration, but there's no self-healing. Consider calling `_ensure_source_graph_node` on *both* branches — the helper is already idempotent, so the cost is one extra SELECT per upsert.

- **Leading-underscore helpers imported across modules** (`backend/scripts/backfill_graph_nodes_for_sources.py:28-32`). `_ensure_source_graph_node`, `_ensure_document_graph_node`, and `_format_source_label` are marked private but imported from the backfill script. Pick a side: either drop the underscores (they're part of the de-facto public API now) or re-export them via non-underscore aliases. Small, but avoids the "private API with external callers" smell.

- **Duplicate regex between backend and frontend** (`backend/app/services/wikilink_parser.py:39` and `frontend/src/lib/remark-wikilinks.ts:22`). `_DOCUMENT_LINK_RE` and `DOCUMENT_LINK_RE` are byte-for-byte equivalent and must stay in sync. The PR also copies the `filing` regex. Not actionable here (Python/TS split is real), but worth a comment on each pointing to its sibling so a future change keeps both in lock-step.

- **`citation_map` threaded through a closure wrapper** (`backend/app/api/routes.py:2517-2525`). The wrapper function now exists only to capture `citation_map` from the enclosing scope. Could be simplified by passing `citation_map` directly to the call sites, letting the closure disappear entirely. Minor — the current form does preserve the existing call-site signatures, which is defensible.

### 🔵 Nitpick

- `cite_translator.py` — `import re` is now top-level but the old closure did it inline; fine, just noting the intentional fix.
- Test fixture reuses `_DOC_UUID` across both parser and translator tests with the same value (`test_wikilink_parser.py:11` / `test_cite_translator.py:4`). Harmless but using distinct constants per file avoids false coupling.
- `document-link` renderer in `notes/page.tsx:489` duplicates ~95% of the `filing-link` renderer right above it. Could be factored into a shared `CitationLink` component, but worth waiting until the third copy appears.
- `makeDocumentLinkNode` default `"Document"` display text (`remark-wikilinks.ts:154`) — the backend tests assert a `"Source"` fallback for unresolvable labels; these aren't the same code path (one is sidecar label, one is node display), but the inconsistency might confuse a future reader. One-line comment would suffice.
- `graph.py:24` inline node_type comment is great — consider promoting it to a `Literal["note","source","document","chart"]` type alias so the frontend `GraphNodeData` union and the backend stay coupled at the type level.

## Positive Feedback

- **Legacy fallback in `sync_edges` is the star of this PR** (`notes_service.py:356-368`). Re-querying unmatched `filing_ref_ids` against `node_type='document'` means historical notes keep working without a content migration. That's exactly the right call — cheap, reversible, and avoids rewriting user-owned markdown.
- **Test `test_cite_translator.py:33` explicitly asserts `"filing:" not in rewritten`** — pinning the invariant that the translator must never emit the legacy shape. Great regression guard for the exact bug this PR fixes.
- **Split of `note_org_filter` vs `public_or_org_filter`** (`notes_service.py:313-322`) is well-commented and gets the authz model right: private notes stay strict, but public sources (filings/documents/charts) are visible to org-scoped notes. The `or_` composition is correct.
- **Backfill script is idempotent and `--dry-run` aware**, with separate seed/resync phases and clear logging. Easy to reason about and re-run if it partially fails.
- **Sidecar writes both `document_id` and `filing_id`** as a legacy alias (`cite_translator.py:62-63`) — the forward/backward-compat dance is handled cleanly in one place.

## Conclusion

**Comment** — leaning toward approve once the race condition in `_ensure_*_graph_node` is either fixed or explicitly acknowledged as acceptable. Everything else is non-blocking polish. The core fix is correct, well-tested, and the legacy fallback strategy is genuinely good engineering.

nvasilevv:
# Follow-up review — fix commit 1a27f8d

Thanks for the quick turnaround. Most of the feedback is fully addressed:

- ✅ **Unique constraint** added on `(node_type, ref_id)` at `graph.py:14-21`, and the backfill script runs `CREATE UNIQUE INDEX IF NOT EXISTS` for pre-existing databases (`backfill_graph_nodes_for_sources.py:131-147`). The "pre-PR rows are only note/chart types with no duplicates" assertion is worth a quick sanity check on staging before the real run — if it holds, the index creation will be instant; if it doesn't, the script fails loudly which is the right failure mode.
- ✅ **`upsert_filing` self-heal** on the existing branch (`repository.py:425-441`) — clean and idempotent.
- ✅ **Helper rename** — `format_source_label`, `ensure_source_graph_node`, `ensure_document_graph_node` are now properly public, and all call sites (including the backfill script) updated.
- ✅ **Cross-reference comments** on the duplicated wikilink regexes in both `wikilink_parser.py:18-22` and `remark-wikilinks.ts:17-21`. Clear and symmetric.

## 🟡 One new concern — transaction scope of the rollback

`_upsert_graph_node` (`repository.py:349-362`) catches `IntegrityError` and calls `await session.rollback()`. The issue: in SQLAlchemy/SQLModel async sessions, `session.rollback()` rolls back the **entire current transaction**, not just the failed INSERT. Look at the call sites:

- `upsert_filing` insert path (`repository.py:456-484`): `session.add(filing)` → `session.flush()` → `ensure_source_graph_node(...)`. If the graph_node race fires, the rollback wipes the just-flushed Filing row too. The caller then commits an empty transaction and returns a `filing` object whose row no longer exists.
- `create_document` (`repository.py:660-688`): same shape, same risk — the Document row gets rolled back alongside the graph_node attempt.

In practice the race window is narrow (two concurrent seeders for the same `ref_id`), so this is unlikely to fire often, but when it does it silently loses the caller's primary write. Recommended fix: wrap the INSERT attempt in a **SAVEPOINT** so the rollback is scoped:

```python
node = GraphNode(org_id=None, node_type=node_type, ref_id=ref_id, label=label)
try:
    async with session.begin_nested():  # SAVEPOINT
        session.add(node)
        await session.flush()
    return node
except IntegrityError:
    # SAVEPOINT already rolled back — outer transaction intact.
    existing = (await session.exec(
        select(GraphNode).where(
            GraphNode.node_type == node_type,
            GraphNode.ref_id == ref_id,
        )
    )).first()
    if existing is None:
        raise
    return existing
```

`begin_nested()` issues a real `SAVEPOINT` on Postgres, so on `IntegrityError` only the graph_node insert is rolled back and the enclosing Filing/Document insert survives. This is the standard SQLAlchemy pattern for "try insert, fall back to fetch on conflict."

(Alternatively, on Postgres you could sidestep the whole dance with `INSERT ... ON CONFLICT (node_type, ref_id) DO NOTHING RETURNING *`, but `begin_nested()` keeps the code dialect-agnostic.)

## Conclusion

**Request changes** on the rollback scoping — it's a real correctness hazard for `upsert_filing` and `create_document`, even if rare. Everything else is solid and ready to merge once that's addressed.

nvasilevv:
# Follow-up review — fix commit 1f5a2fc

SAVEPOINT fix looks correct. `async with session.begin_nested()` scopes the rollback to just the graph_node INSERT attempt, so an `IntegrityError` from the unique constraint no longer unwinds the enclosing Filing/Document transaction. The re-fetch after the `except` block runs in the still-alive outer transaction, which is exactly the right shape.

The docstring update is clear about *why* SAVEPOINT is needed, which future-proofs the pattern against someone "simplifying" it back.

## Conclusion

**Approve.** All five items from the original review are now addressed cleanly:

- ✅ Race condition — UniqueConstraint + SAVEPOINT-scoped upsert
- ✅ `upsert_filing` self-heal on existing branch
- ✅ Public helper renames (`format_source_label`, `ensure_*_graph_node`)
- ✅ Cross-reference comments on duplicated wikilink regexes
- ✅ Backfill script seeds the unique index for pre-existing DBs

Ready to merge once the deploy-plan checklist (dry-run → real backfill → visual verification of the Games Operators note) passes on staging.

PR #13, opened 2026-04-13
Add optimistic concurrency to update_note via base_revision

nvasilevv:
## Summary
- Agent (and REST) `update_note` now accepts `base_revision` + `force`, rejecting writes when the note has advanced past the pinned revision so concurrent edits aren't silently clobbered.
- The agent executor returns a structured `revision_conflict` error with `current_revision`, `base_revision`, and a re-read/re-merge/retry hint — the model can react in-loop.
- The REST PUT maps the same conflict to HTTP 409.

## Why
Prompted by a real incident: an agent chained `read_note(revision=2)` → `update_note(..., revision_number=...)`, hallucinating a kwarg that didn't exist on the schema and crashing with `TypeError`. Making the concept first-class removes the footgun and gives the agent an actual mechanism for the thing it was reaching for. The check only fires when `content` is supplied; title-only updates are untouched.

## Test plan
- [ ] Unit: `update_note` with stale `base_revision` raises `NoteRevisionConflictError`; with matching `base_revision` proceeds; with `force=True` bypasses the check.
- [ ] Agent path: call `execute_update_note` with a stale `base_revision` and confirm the returned dict has `error: "revision_conflict"` and the hint.
- [ ] REST: `PUT /api/notes/{id}` with a stale `base_revision` returns HTTP 409 with the conflict body.
- [ ] Smoke: existing update flows (no `base_revision` passed) still behave identically.


nvasilevv:
## Senior Staff / Security Review — PR #13

### Summary
Adds optimistic-concurrency (`base_revision` + `force`) to `update_note` across the service, agent executor, REST PUT, and the agent tool schema, with a dedicated `NoteRevisionConflictError` mapped to HTTP 409 and a structured agent-facing error payload that includes a re-read/re-merge/retry hint. The diff also bundles an unrelated backfill-script fix (`_resync_note_edges` rework + `Organization` import for ORM registration).

### Review Findings

**🔴 Critical / Security**
- **Race still reachable under the check.** `backend/app/services/notes_service.py:238-249` reads `max(revision_number)` and then inserts a new row with `max_rev + 1` in the same transaction, but nothing locks the note or the revisions table. Two concurrent `update_note` calls with the same `base_revision=N` can both observe `max_rev=N`, both pass the guard, and both attempt to insert revision `N+1`. Whether this corrupts history depends on whether `note_revision` has a unique constraint on `(note_id, revision_number)` — if it does, one caller gets a raw `IntegrityError` (not a clean `revision_conflict`); if it doesn't, you silently end up with two “latest” revisions and `get_latest_revision` picks arbitrarily. Either way the invariant the PR is selling isn't actually enforced. Fix: either `SELECT … FOR UPDATE` the `Note` row at the top of `update_note`, or wrap the check+insert in `with_for_update()` against the max-revision query, and ensure the unique constraint exists so the DB is the source of truth.

**🟡 Improvement**
- **Scope bleed.** The PR description and title are exclusively about `base_revision`, but the diff also contains the `backfill_graph_nodes_for_sources.py` rework (per-note sessions, raw SQL `SELECT id FROM note`, `Organization` side-effect import). That change is independently reasonable — and the `Organization` comment is genuinely useful — but it should land as its own PR so a reviewer looking at optimistic concurrency isn't also auditing ORM-session lifetime semantics. At minimum, call it out explicitly in the description.
- **Title-only updates silently ignore `base_revision`.** `notes_service.py:237` only runs the guard inside the `content is not None` branch. An agent that passes `base_revision` alongside a title-only update gets no error and no check — the kwarg is accepted and dropped. The tool schema docstring mentions this (“Only checked when content is supplied”), but from the agent's perspective passing a pin and having it quietly ignored is exactly the footgun the PR is trying to remove. Consider either (a) raising `ValueError` if `base_revision` is passed without `content`, or (b) running the check unconditionally when pinned — title collisions are rare but cheap to detect.
- **Inconsistent agent error shape.** `execute_update_note` returns `{\"error\": \"revision_conflict\", …}` for conflicts but `{\"error\": str(exc)}` (a human string) for `NoteNotFoundError`. The model now has to branch on `error` being either a machine code or a sentence. Normalize to `{\"error\": \"not_found\", \"message\": str(exc)}` so the loop can dispatch on a stable set of codes.
- **`force=True` has no audit trail.** The only log line is `update_note called: … force=%r` at INFO. Given `force` intentionally overwrites another writer's work, consider logging at WARNING when `force=True` *and* a real conflict was bypassed (i.e. `base_revision is not None and max_rev != base_revision and force`). Otherwise incident forensics — the exact scenario that motivated this PR — has nothing to grep for.
- **Missing test coverage.** The PR description's test plan checkboxes are all unchecked and no test files are touched in the diff. For a change whose whole value proposition is correctness under concurrency, the unit + agent-path + REST 409 cases called out in the description should actually exist before merge. The race case above is also worth a dedicated test (two concurrent tasks, both pinned to the same revision).

**🔵 Nitpick**
- `logger.info(\"update_note called: note_id=%r base_revision=%r force=%r\", note_id, base_revision, force)` — the existing log style in this file typically also includes `email`/caller; worth matching for consistency.
- `NoteRevisionConflictError` could subclass a common `NotesServiceError` base so future service-layer exceptions have a single catch point, but there isn't one today so this is purely forward-looking.
- In the backfill script, `text(\"SELECT id FROM note\")` works but `select(Note.id)` keeps the query typed and survives a table rename — the comment already explains *why* a raw query is needed (expiration), which is fine; the `SELECT id` itself doesn't need to be raw.
- `base_revision: int | None = None` in the REST `NoteUpdate` model has no validation that it's ≥ 1; a client passing `0` or a negative number will just get a conflict with a slightly nonsensical message. A `Field(ge=1)` is one line.

### Positive Feedback
- The structured agent error with `current_revision`, `base_revision`, **and** an explicit natural-language hint telling the model to re-read → re-merge → retry is the right shape — it turns a failure mode into a recoverable in-loop action instead of a dead end. Explicitly mentioning `force=true` in the hint and bounding when to use it is also well-considered.
- The decision to only guard `content` updates (and document it) is principled — title changes don't invalidate the revision chain, so pinning them would be noise.
- Adding the `update_note` tool-schema description to prescribe “always read_note first and pass revision_number as base_revision” turns the schema itself into the documentation the agent reads, which is exactly the lever that prevents the original `TypeError` incident from recurring.
- Preserving the prior `citations_json` when the caller doesn't supply one (unchanged in this PR but visible in context) continues to be the right default and isn't disrupted by the new code path.

### Conclusion
**Request Changes** — the optimistic-concurrency contract isn't actually enforced against concurrent writers without a row lock (or a DB unique constraint + conflict translation), and the PR ships without the tests its own description promises. Split the backfill-script fix into a separate PR, add the lock + tests, normalize the agent error shape, and this is a clean merge.

— Reviewed per the engineering/PR reviewer rubric.

nvasilevv:
## Follow-up Review — addressing commits `9ff762d` + `ccc047e`

Walked the two new commits against each finding from my earlier review.

### Validation vs. original review

**🔴 Race on `max_rev` read (Critical) — ✅ resolved.**
`notes_service.py` now does `SELECT … FOR UPDATE` on the `Note` row before reading `max(revision_number)`, so two concurrent pinned writers serialize cleanly. Belt-and-braces: the `uq_note_revision(note_id, revision_number)` unique constraint genuinely exists in `models/note_revision.py`, **and** `session.commit()` is wrapped in `try/except IntegrityError` that rolls back, re-reads `max_rev`, and re-raises as `NoteRevisionConflictError`. So even backends that ignore `FOR UPDATE` (or a row that somehow skips the lock path) still surface a clean 409, not a 500. This is stronger than what I asked for.
- Minor: the locked path re-implements `get_note`'s authz inline rather than threading a `for_update=True` flag through `get_note`. Works, but it's the only place in the file with duplicated visibility logic — worth a small refactor later.

**🟡 Title-only updates silently ignore `base_revision` — ✅ resolved.**
`update_note` now raises `ValueError(\"base_revision is only meaningful when content is supplied\")` up-front, REST maps it to HTTP 400, and the agent executor maps it to `{\"error\": \"invalid_argument\", …}`. Exactly the loud-failure shape.

**🟡 Inconsistent agent error shape — ✅ resolved.**
Both `execute_read_note` and `execute_update_note` now return `{\"error\": \"<code>\", \"message\": <str>}` for `not_found` / `invalid_argument` / `revision_conflict`. The model can dispatch on a stable set of codes.

**🟡 `force=True` has no audit trail — ✅ resolved.**
The WARNING only fires when `force=True` **and** `max_rev != base_revision` — i.e. when force actually bypassed a real conflict. Message includes `note_id`, `email`, `current_revision`, `base_revision`. That's exactly the forensic anchor, and it correctly doesn't spam when force is a no-op.

**🟡 Missing tests — ✅ partially resolved.**
`tests/test_notes_service_concurrency.py` covers: stale `base_revision` raises, matching `base_revision` passes, `force=True` logs a WARNING only on real conflict, `base_revision` without content is `ValueError`, and Pydantic rejects `base_revision < 1`. The test doc explicitly states the commit/refresh happy path and REST 409 dispatch are **manual smoke** rather than mocked, with the reasoning \"mocking the whole commit/refresh chain would test the mocks, not the invariant\" — that's a defensible call.
- **Gap:** the race itself has no test. The whole point of the `FOR UPDATE` + `IntegrityError` fallback is that two concurrent writers pinned to the same revision can't both win, and that invariant is not exercised in CI. It needs a real DB (two sessions, two transactions) which is why it wasn't written, but it's also the single most important case. Not a blocker given the belt-and-braces design, but the DB-level guarantees are effectively unverified until someone adds an integration test.

**🟡 Scope bleed (backfill changes in a concurrency PR) — ❌ not addressed.**
The backfill commit is still in the PR. The author tightened the id-collection query to `select(Note.id)` (per my nitpick) but didn't split the work out. Non-blocking — it was an Improvement, not Critical — but the observation stands for next time.

**🔵 Nitpicks — all resolved.**
- `email` now included in the `update_note` INFO log line.
- `Field(default=None, ge=1)` on `NoteUpdate.base_revision`, with `ValueError` → HTTP 400.
- `select(Note.id)` replaces the raw `text(\"SELECT id FROM note\")` in the backfill script; the comment explaining *why* a short-lived session is needed was retained.

### Conclusion
Every critical and improvement item I flagged is addressed in the code with thoughtful choices. The `FOR UPDATE` + unique-constraint + `IntegrityError` translation is a stronger design than my original ask. Two residual gaps: (1) no CI-level race test against a real DB — design-mitigated but unverified, (2) the backfill-script fix is still bundled.

Moving verdict from **Request Changes** to **Approve with comments**. Suggest a follow-up PR to add the two-session integration test for the race path so the DB-level guarantee has a regression anchor.

PR #14, opened 2026-04-13
Document Viewer V2: TipTap editor + theme_config

nvasilevv:
## Summary
- Phase 1+2: Replace the chat-side document viewer's `react-markdown` renderer with TipTap (read-only). Adds `TipTapViewer` + custom nodes (`FilingLink`, `DocumentLink`, `NoteLink`, `ChartEmbed`) with HTML-bridge parsing and `toMarkdown` serializers.
- Phase 3: Replace the research notes textarea + preview pane with a single WYSIWYG TipTap editor. Wikilink autocomplete is rewired to ProseMirror (`state.doc.textBetween` + `view.coordsAtPos`). Adds per-note `theme_config` (font family + size) stored as a JSONB column on `Note`, applied as inline styles on a wrapper outside the editor so markdown stays clean.
- Threads `base_revision` through the frontend `updateNote()` call so the notes editor participates in the optimistic-concurrency contract.

Note: this branch also carries the `note-base-revision` backend hardening commits (row-lock on the concurrency path, `base_revision >= 1` validation, `IntegrityError` → revision-conflict translation). Those predate my work here and came along via a merge.

## Test plan
- [ ] Open a chat citation → the `DocumentViewer` renders the markdown via TipTap. Tables, headings, lists, and `[[filing:…]]` / `![[chart:…]]` wikilinks all render; Ctrl+F search and citation auto-highlight still work.
- [ ] Open a research note → the editor loads in view mode. Click into Edit → TipTap becomes editable.
- [ ] Type `[[` in the editor → wikilink autocomplete opens at the caret; selecting a filing/note inserts the node and serializes back to `[[filing:UUID|Title]]` / `[[note_id:UUID|Title]]` on save.
- [ ] Type `![[` → chart-only autocomplete; picking a chart inserts a block-level embed that renders the live `ChartRenderer`.
- [ ] Click a filing-link / document-link inside a note → opens the `DocumentViewer` slide-in with citation highlight.
- [ ] Change the font family / size dropdowns while editing, save → reload the note and verify `theme_config` is persisted and re-applied.
- [ ] Save a note where another tab has advanced the revision → backend returns 409 revision_conflict (UI surfacing of the conflict can be a follow-up).
- [ ] `npm run build` passes; `tsc --noEmit` clean.


nvasilevv:
## Review: Document Viewer V2 — TipTap editor + theme_config

### Summary
Swaps the chat document viewer and research-notes editor from `react-markdown` + textarea to a unified TipTap-based viewer/editor, with custom nodes for vault wikilinks (`filing-link`, `document-link`, `note-link`, `chart-embed`). Adds a `theme_config` JSONB column on `note` for per-note font family/size and threads `base_revision` into the frontend save path.

### Findings

**🔴 Critical / Security**

- **`wikilinksToHtml` runs unscoped over raw markdown** (`frontend/src/lib/tiptap/wikilink-bridge.ts`). It rewrites wikilinks even inside fenced code blocks and inline code. A user who writes `` `[[filing:…]]` `` as a literal example will have it silently replaced with HTML before `tiptap-markdown` parses it — so it renders as a live link instead of literal text. Not exploitable (IDs are UUID-constrained, display text is escaped), but a round-trip correctness bug that corrupts user input. Tokenize code spans/fences first, or move this to a micromark extension the way `remark-wikilinks` did.
- **`attr()` escape lists are incomplete but safe given current usage.** Both `wikilink-bridge.ts` and `notes/page.tsx` escape `&`/`"`/`<` only. Fine for `"`-quoted attributes today, but one refactor away from a hole (e.g. switching to single-quoted attrs, or embedding into text content). Pin the assumption with a comment, or escape `>` and `'` too.

**🟡 Improvement**

- **Separate `updateNote` PUT to persist `theme_config` on create** (`notes/page.tsx` `save`): on "save new note" you call `createNote({title, content})` then immediately `updateNote(created.id, {theme_config})`. Two round trips, a window where the note exists without its theme, and no `base_revision` on the follow-up. Extend `createNote` to accept `theme_config` instead.
- **`onEditorReady` effect churns.** The parent passes `editing ? setEditor : undefined`, which changes identity every render. The effect in `tiptap-viewer.tsx` re-runs each time. Harmless but noisy — memoize the callback or guard with a ref.
- **`wikilinksToHtml` not memoized** in `TipTapViewer`, and the "reload when content changes" effect calls `storage.markdown.getMarkdown()` on every keystroke in editable mode. Skip the effect entirely when the parent owns `content` (i.e. when `onChange` is wired).
- **Display-text round-trip bug** in `lib/tiptap/nodes.tsx` serializers: `displayText !== "Filing"` / `"Document"` / `"Note"` is used to decide whether to emit the `|display` suffix. If a user's actual label literally equals `"Filing"`, it silently drops. Track whether display text was explicit via a separate attribute instead of magic-string comparison.
- **Non-embed chart wikilinks regress to plain text.** `wikilink-bridge.ts` only has a regex for `![[chart:UUID]]` (embed). The pre-TipTap path also handled `[[chart:UUID|Title]]` as a styled reference link, and the autocomplete code still emits that syntax for the non-embed path. Add a `ChartLinkNode` or extend the chart regex.
- **Code duplication** in `handleAutocompleteSelect` — four near-identical `chain().focus().deleteRange().insertContent().run()` calls. Factor into a single helper taking the HTML string.
- **`document` shadowing** in `tiptap-viewer.tsx` `handleClick` — `const document = target.closest(...)` shadows the global. Rename to `docEl`.
- **`init_db.py` raw-SQL migration is repo-consistent**, but `theme_config` has no default and no index. Fine for now; flag if you ever want to filter by it.

**🔵 Nitpick**

- `frontend/tsconfig.tsbuildinfo` shows up as untracked in `git status` — add to `.gitignore`.
- PR body calls out "UI surfacing of the conflict can be a follow-up" for 409 handling — make sure there's a ticket.
- Markdown getter does `editor.storage as unknown as {markdown: {getMarkdown(): string}}` twice — pull out a small typed helper.
- `FONT_FAMILIES` CSS strings live in the component; if reused elsewhere (print/export?) move to a shared module.
- `handleClick` repeats `target.closest` three times — iterate a small `[selector, kind]` array.

### Positive
- Clean separation: headless `TipTapViewer` serves both the read-only chat viewer and the editable notes page from one component — exactly the right direction.
- Using `view.coordsAtPos` + `view.dom.getBoundingClientRect` to position the autocomplete is a big simplification over the old mirror-div caret-measurement hack.
- Custom TipTap nodes are well-scoped atoms with clear parse/render/serialize triples, and the `wikilink-bridge` → `parseHTML` → `toMarkdown` round-trip contract is well documented.
- Event delegation via `data-*` attributes on the wrapping `div` instead of per-node React handlers is the right pattern for TipTap NodeViews.
- `base_revision` threading into `updateNote` is a nice follow-through on the concurrency work from #13.

### Conclusion
**Request Changes** — primarily because of the code-block round-trip bug in `wikilinksToHtml` (data corruption on user input) and the display-text magic-string comparison. The other items are improvements, not blockers. Architectural direction is right.

nvasilevv:
### Re-review after 375dfcd

All items from my previous review are addressed. Quick verification:

**🔴 Critical — both fixed**
- `wikilinksToHtml` now iterates `PROTECTED_RE` matches (triple-backtick fences, tildes, single-backtick inline) and only transforms segments outside code regions. Literal `` `[[filing:…]]` `` in notes is preserved. Documented caveat that double-backtick inline code isn't handled — acceptable.
- `attr()` escape assumption now has a docstring pinning it to double-quoted attribute usage.

**🟡 Improvements — all addressed**
- `createNote` extended end-to-end (backend + frontend); the second-PUT workaround in `save()` is gone.
- `onEditorReady` stashed in a ref; the editor-ready effect depends only on `[editor]`, so parent re-renders no longer churn it.
- Content-reload effect early-returns when `onChange` is wired — editable mode no longer does a full `getMarkdown` on every keystroke.
- Display-text round-trip: all four link nodes default `displayText` to `null`, serializers use `!= null` instead of magic-string comparison, and the bridge omits `display-text` entirely when no `|display` was in source. A label equal to "Filing"/"Document"/"Note" now round-trips cleanly.
- `ChartLinkNode` added with its own parseHTML/renderHTML/serialize + `CHART_LINK_RE` in the bridge. Autocomplete now inserts `<chart-link>` HTML for the non-embed chart case instead of plaintext `[[chart:…]]`.
- `insertHtml` helper factored out of `handleAutocompleteSelect`.
- `document` shadowing in `handleClick` fixed via the `NODE_CLICK_SELECTORS` table and a `docEl` local.
- Typed `getMarkdown(editor)` helper extracted and reused.

**🔵 Nits**
- `*.tsbuildinfo` added to `.gitignore`.
- `FONT_FAMILIES` relocation wasn't done — minor, non-blocking.

**Follow-up filed**
- Clicking an inline `chart-link` is currently a no-op because `NODE_CLICK_SELECTORS` doesn't include `[data-chart-link-id]` and `ClickedVaultNode` has no `"chart"` variant. Non-regression relative to pre-TipTap behavior, so not a blocker. Tracked separately.

### Conclusion
**Approve**. Nice, thorough response to the review.

PR #15, opened 2026-04-13
Centralize ORM model registration in app.models

nvasilevv:
## Summary
- `backend/app/workers/worker_queue.py` (the backfill/realtime worker entrypoint) never imported `app.models.organization` and never called `init_db()`. Its metadata came transitively from `repository.py`, which imports `GraphNode` but not `Organization`.
- At flush time in `_upsert_graph_node` the worker hit `NoReferencedTableError: Foreign key associated with column 'graph_node.org_id' could not find table 'organization'`, so **every** filing processed by the backfill worker has been failing since the knowledge-graph PR landed. ~358 messages are currently stuck in `/tmp/ceewire-backfill-queue` on prod.
- The API container was unaffected because it calls `init_db()` on startup, which imports every model explicitly.

## Fix
Make `backend/app/models/__init__.py` the single source of truth for model registration. Importing the package (or any submodule, since Python evaluates the parent `__init__` first) now registers all 47 tables with `SQLModel.metadata`.

- `backend/app/models/__init__.py` — imports every table model; exposes `HAS_BRAIN_ENTRY` for pgvector-optional environments.
- `backend/app/db/init_db.py` — 35 lines of per-model imports collapsed into `import app.models`.
- Every worker `__main__` entrypoint (`worker_queue`, `listener`, `worker_download`, `ir_crawler`, `gpw_poller`, `worker_bse_eod`, `eacsp_poller`) declares `import app.models  # noqa: F401` near the top. Technically redundant once `app/models/__init__.py` is populated (any `from app.models.X` submodule import triggers it), but kept as an explicit marker so future maintainers can see the FK-resolution dependency at the entrypoint level.

No model module imports from `app.services` / `app.workers` / `app.api` (verified with grep), so there is no circular-import risk from populating the package `__init__`.

## Verification
Copied the new `app/models/__init__.py` into the live `ceewire-backfill-worker` container and ran:

```
import app.models
from sqlmodel import SQLModel
from sqlalchemy.orm import configure_mappers
configure_mappers()
```

Result: `47 tables`, `organization: True`, `graph_node: True`, mappers configured with no `NoReferencedTableError`. Container reverted afterwards.

## Test plan
- [ ] Rebuild and redeploy `ceewire-backfill-worker` and `ceewire-worker`
- [ ] Confirm the ~358 queued messages drain and land as `completed` filings rather than failing at `_upsert_graph_node`
- [ ] Check `ceewire-listener`, `ceewire-gpw-poller`, `ceewire-eacsp-poller`, `ceewire-ir-crawler`, `ceewire-bse-eod` all start cleanly with the new import
- [ ] `ceewire-api` still boots via `init_db()` (functional equivalence of the collapsed import list)


nvasilevv:
## Review: PR #15 — Centralize ORM model registration in app.models

### Summary
Fixes a production outage where `ceewire-backfill-worker` crashed on every filing with `NoReferencedTableError` because its entrypoint never registered `Organization` with SQLModel metadata. The fix makes `app/models/__init__.py` the single source of truth for model registration, collapses `init_db.py`'s 35-line import block, and adds an `import app.models` marker to every worker entrypoint plus a regression test.

### Review Findings

🔴 **Critical/Security** — None.

🟡 **Improvement**
- **Redundant `import app.models` markers in workers.** The PR body already acknowledges these are technically redundant (any `from app.models.X` in the worker's transitive imports triggers `__init__.py`). The "explicit marker" rationale is reasonable, but it creates a second place future maintainers can forget to update. Consider instead moving the guarantee into `app/db/session.py` (`get_session_context` / `engine` module), so *anything* that touches the DB picks it up automatically — workers then don't need per-file markers at all. Lower-risk alternative: keep the markers but add a lint/test that fails if a worker imports `app.db.session` without also importing `app.models`.
- **`init_db.py` still exposes `_HAS_PGVECTOR` via an underscore-prefixed alias re-import.** `from app.models import HAS_BRAIN_ENTRY as _HAS_PGVECTOR` + `if not _HAS_PGVECTOR:` logs the warning at module-import time (side effect on `init_db` import), whereas previously the warning was logged inside the `try/except`. Functionally equivalent today, but the warning now fires every time `init_db` is imported rather than once at the site where the decision is made. Minor; move the `logger.warning` to `app/models/__init__.py` next to the `except ImportError` so it logs once, at the actual source of truth.
- **`ChatCompaction` appears in `__init__.py` but was not in the old `init_db.py` import list.** Worth confirming it was already registered transitively (via `chat.py`) — if so, fine; if not, this PR is silently adding a new table to `metadata.create_all()` callers, which could matter for test DB fixtures. (The regression test doesn't cover it.)
- **Regression test coverage is narrow.** `test_importing_app_models_registers_full_metadata` checks 5 representative tables. Given the whole point of this PR is "we can't trust the import list to stay in sync," a stronger test would assert `len(SQLModel.metadata.tables) >= N` or iterate over every class in `app.models.__all__` and assert its `__tablename__` is registered. Otherwise a future contributor can add a model to `__init__.py` but forget `__all__`, or vice versa, and the test still passes.
- **`__all__` includes `HAS_BRAIN_ENTRY`** (a bool) alongside classes. Not wrong, but unusual — `__all__` is conventionally type/function exports. Consider a separate module-level constant export convention or just dropping it from `__all__` since nothing `from app.models import *` would meaningfully want it.

🔵 **Nitpick**
- Comment in `init_db.py` says "do NOT add per-model imports here" — good. Mirror the same warning as a comment at the top of `app/models/__init__.py` so the rule is discoverable from the file it actually governs.
- The `try/except ImportError` for `BrainEntry` swallows the warning log in the new location. If pgvector is genuinely missing in prod, operators will no longer see the log line unless `init_db` is imported. Acceptable, but note it.
- Worker marker comments are slightly inconsistent (`ir_crawler.py` lost its `Filing` import with comment "needed for FK resolution"; the new marker says "register all ORM models"). Fine, just noting the rationale has shifted from "this specific FK" to "everything."

### Positive Feedback
- Root-cause analysis in the PR description is exemplary: identifies the exact failure mode (`_upsert_graph_node` → `NoReferencedTableError`), explains why API was unaffected (calls `init_db()`), quantifies impact (~358 stuck messages), and verifies the fix in the live container before shipping.
- The regression test's second case (`test_graph_node_org_fk_resolves`) is exactly the right shape — it tests the *specific invariant* that broke, not a generic smoke test. This is the kind of test that actually prevents a repeat incident.
- Collapsing `init_db.py` from 35 scattered imports to one is a meaningful maintainability win. The old pattern was an accident waiting to happen — this PR not only fixes the bug, it removes the class of bug.
- Verified absence of circular imports with grep before populating the package `__init__` — good diligence, since that's the usual reason people avoid this pattern.

### Conclusion
**Approve** — with a suggestion to strengthen the regression test (assert all `__all__` entries are registered) and consider moving the pgvector `logger.warning` into `app/models/__init__.py` so the warning fires at the source of truth. The worker-level `import app.models` markers are debatable but defensible; I'd accept them as-is given the recent outage makes the "explicit is better than implicit" argument more persuasive than usual.

Issue #17, opened 2026-04-13
Document Viewer V2: remaining work after PR #14


nvasilevv:
Audit of the Document Viewer V2 plan (see `ceewire_m/DocumentViewer V2.md`) against `main` after PR #14 (`368ca3d`). The core of Phases 1–3 is in, but a few items from the plan are still outstanding.

## Implemented on main

- TipTap deps installed: `@tiptap/react`, `@tiptap/starter-kit`, `@tiptap/suggestion`, `@tiptap/pm`, `tiptap-markdown`, `tippy.js`.
- `<TipTapViewer>` at `frontend/src/components/tiptap/tiptap-viewer.tsx` — unified read-only + editable component with markdown round-trip via `tiptap-markdown` plus custom vault-node serializers in `frontend/src/lib/tiptap/nodes.tsx`. Wikilink→HTML bridge at `frontend/src/lib/tiptap/wikilink-bridge.ts`.
- Phase 2 chat viewer swap — `frontend/src/components/feed/document-viewer.tsx` now renders via `<TipTapViewer readOnly />`.
- Phase 3 notes editor — `frontend/src/app/[locale]/notes/page.tsx` replaces the textarea with TipTap; font-family + font-size dropdowns wired outside the editor and applied as inline style on the wrapper.
- Backend `theme_config` JSONB column on `Note` (`backend/app/models/note.py`), plumbed through `NoteCreate` **and** `NoteUpdate` (`backend/app/api/notes.py`) and `notes_service.create_note`.
- API client + types updated (`frontend/src/lib/api.ts`, `frontend/src/lib/types.ts`).
- Click-through for vault nodes preserved via event delegation on `data-filing-id` / `data-document-id` / `data-note-id`.

### 1. Alembic migration for `Note.theme_config`
The column is declared on the SQLModel and created via `backend/app/db/init_db.py`, but there is no dedicated Alembic migration in PR #14. Fresh DBs are fine; existing prod/staging DBs will need a manual `ALTER TABLE notes ADD COLUMN theme_config JSONB NULL;` or a proper migration.

### 2. Slash-command menu (Phase 3, step 8)
`@tiptap/suggestion` and `tippy.js` are installed, but `tiptap-viewer.tsx` only registers `StarterKit`, `Markdown`, and `VaultWikilinkNodes` — no suggestion extension is wired. The `/chart`, `/cite`, `/filing` quick-insert menu from the plan is not implemented yet.

[...]

PR #18, opened 2026-04-13
Add slash-command menu to TipTap notes editor

nvasilevv:
## Summary

- New `SlashCommandExtension` (`frontend/src/lib/tiptap/slash-command.tsx`) wrapping `@tiptap/suggestion` with a `tippy.js` popup rendered via `@tiptap/react`'s `ReactRenderer`.
- Four commands: `/note`, `/filing`, `/chart`, `/embed-chart`. Selecting one deletes the `/…` range and inserts the matching wikilink prefix (`[[filing:`, etc.), which the existing `detectAutocomplete()` listener in `notes/page.tsx` picks up to open `WikilinkAutocomplete` — no duplicate search logic, no backend changes.
- Registered in `tiptap-viewer.tsx` only when `!readOnly`, so the chat citation viewer is unaffected.

Addresses the outstanding slash-command item from #17. (During planning we confirmed the other #17 items — Alembic migration, WikilinkAutocomplete wiring, mark.js Ctrl+F — are already shipped.)

## Test plan

- [ ] `cd frontend && npm run dev`, open a research note
- [ ] Type `/` on a blank line → popup shows 4 commands
- [ ] Type `/fil` → filtered to "filing"; press Enter → `[[filing:` inserted and `WikilinkAutocomplete` opens
- [ ] Pick a filing, save, reload → filing link survives the markdown round-trip
- [ ] Repeat for `/note`, `/chart`, `/embed-chart`
- [ ] Negative: `foo/bar` mid-word does not open the popup
- [ ] Negative: `/` inside the chat document viewer (readOnly) does not open the popup
- [ ] Escape dismisses the popup without inserting
- [ ] Regression: Ctrl+F (mark.js) still highlights in the chat document viewer


nvasilevv:
## PR Review — Add slash-command menu to TipTap notes editor

### Summary
Adds a `@tiptap/suggestion`-based slash menu (`/note`, `/filing`, `/chart`, `/embed-chart`) that inserts a wikilink prefix string, intending to hand off to the existing `detectAutocomplete()` → `WikilinkAutocomplete` flow. Only mounted when `!readOnly`. The extension itself is well-structured, but the hand-off contract with `detectAutocomplete` appears broken.

### Review Findings

**🔴 Critical — broken hand-off to `WikilinkAutocomplete` (likely dead UX)**
Selecting `/filing` inserts literally `[[filing:` into the document. `detectAutocomplete` in `frontend/src/app/[locale]/notes/page.tsx:131` then computes `between = text.slice(lastOpen + 2)` → `"filing:"`, and sets `autocomplete.query = "filing:"` with `triggerType: "all"` (it only switches to `"chart"` when preceded by `!`). `WikilinkAutocomplete` passes that straight to `searchReferences("filing:", undefined)`, which in `backend/app/api/notes.py:210` does `q.strip()` and LIKE-matches against titles/tickers — so the backend searches for the literal substring `"filing:"` across all types. Same problem for `/note` (query `"note_id:"`) and `/chart` / `/embed-chart` (query `"chart:"`). Net effect:
- The type is **not** actually filtered (`filterTypes` is never set).
- The user sees "No results found" until they manually delete the literal `filing:` / `note_id:` / `chart:` prefix — which defeats the whole point of the slash menu.

The test-plan step "Type `/fil` → Enter → pick a filing" almost certainly does not work as written. Two reasonable fixes:

1. Teach `detectAutocomplete` to recognize the `note_id:` / `filing:` / `chart:` prefixes, strip them from `query`, and set `triggerType` accordingly, then have `WikilinkAutocomplete` receive a `filterTypes` prop. This keeps the "insert a marker" design.
2. Or simpler: have each slash command insert just `[[` (or `![[` for embed) and pre-open the autocomplete with a `filterTypes` override via editor storage / a shared signal.

Option 2 is less invasive and matches how the autocomplete already works.

**🟡 Improvement — `readOnly` guard doesn't react to prop changes**
`tiptap-viewer.tsx:106` conditionally includes `SlashCommandExtension` inside `useEditor`, but TipTap doesn't recreate the editor when extensions change. If a caller toggles `readOnly` at runtime the slash extension stays in whichever state it had at mount. In practice the two current callsites (notes page, chat viewer) are fixed, so this is latent — worth a one-line comment or a `readOnly` entry in the effect deps list depending on how the viewer is used.

**🟡 Improvement — `SLASH_COMMANDS` duplicates knowledge already encoded in `nodes.tsx`**
The serialization prefixes (`[[note_id:`, `[[filing:`, `[[chart:`, `![[chart:`) are defined in both `frontend/src/lib/tiptap/nodes.tsx` and now here. If the vault format ever changes they have to be updated in two places. Consider exporting a `WIKILINK_PREFIXES` constant from `nodes.tsx` (or `wikilink-bridge.ts`) and reusing it.

**🟡 Improvement — `allowedPrefixes: [" "]` is narrower than it reads**
With `startOfLine: false` and `allowedPrefixes: [" "]`, `/` after a newline inside a block still works (Suggestion treats block boundaries as prefixes), but typing `/` immediately after punctuation like `)/` or `./` also triggers the menu. Minor, but worth considering `allowedPrefixes: [" ", "\n"]` or validating against a regex if noise becomes an issue.

**🔵 Nitpick — `useEffect(() => setSelectedIndex(0), [items])`**
`items` is a new array every keystroke (from the filter), so this fires on every render. Works, but `[items.length]` or `[props.query]` is cheaper and avoids a harmless redundant state set.

**🔵 Nitpick — `popup` typed as `TippyInstance[] | null`**
`tippy("body", ...)` does return an array, but the code only ever uses `popup[0]`. Using a single instance (`popup: TippyInstance | null` with `popup = tippy("body", ...)[0]`) makes intent clearer.

**🔵 Nitpick — label "note" masks the fact that it's a note link**
In the menu, `/note` and `/filing` look peer-level, but `note_id:` is an internal serialization prefix and was not previously user-facing. If option 2 above is adopted this goes away; otherwise consider a `/research-note` label or at least a clearer description.

### Positive
- Clean separation: the extension is self-contained in one file, keyboard/mouse nav handled correctly, tippy lifecycle (`onStart` / `onUpdate` / `onExit`) cleaned up properly. No XSS risk — `insertContent(props.trigger)` inserts plain text, not HTML.
- Correctly gated behind `!readOnly` so the chat citation viewer stays read-only.
- No backend changes, no new runtime deps beyond what TipTap already pulls in (verify `@tiptap/suggestion` and `tippy.js` are in `frontend/package.json` — they should already be there for the wikilink node views, but worth a glance).

### Conclusion
**Request Changes.** The critical issue is that the selected slash command currently lands the user on a search query that returns nothing relevant — the intended `/filing` → filtered filing picker flow isn't actually wired up. Once that's resolved (either by updating `detectAutocomplete` to understand the prefixes, or by inserting just `[[` and passing a type filter some other way), the rest is polish.

nvasilevv:
## Re-review of a828f30

The new commit addresses the review cleanly. Verified against each finding:

**🔴 Critical — hand-off bug**: **Fixed.** `detectAutocomplete` (`frontend/src/app/[locale]/notes/page.tsx:152-167`) now loops over the new `WIKILINK_BODY_PREFIXES`, strips the matched prefix from `query`, and sets `triggerType` to `"note"` / `"filing"` / `"chart"`. The `filterTypes` prop passed to `WikilinkAutocomplete` is now `[triggerType]` (line 470-473). So `/filing` → `[[filing:` → autocomplete opens with empty query and `filterTypes: ["filing"]`, which is exactly the intended flow.

**🟡 `readOnly` guard**: **Addressed** with an explanatory comment in `tiptap-viewer.tsx` noting the mount-time gate and that both current callers mount with a fixed value. Reasonable — matches reality without adding speculative machinery.

**🟡 Prefix duplication**: **Fixed.** `WIKILINK_PREFIXES` and `WIKILINK_BODY_PREFIXES` are now exported from `nodes.tsx` and consumed by both `slash-command.tsx` and `notes/page.tsx`. Single source of truth.

**🟡 `/chart` vs `/embed-chart` conflation (new wrinkle from the fix)**: Cleanly handled by splitting out an `isEmbed` flag on `AutocompleteState`. Both commands end up with `triggerType: "chart"`, but `handleAutocompleteSelect` now keys the block-vs-inline decision off `autocomplete.isEmbed` (line 216) rather than `triggerType === "chart"`. Good call — that would've been a subtle regression otherwise.

**🔵 `useEffect([items])`**: **Fixed** → `[items.length]`.

**🔵 `popup: TippyInstance[]`**: **Fixed** → unwraps to a single `TippyInstance` at `tippy(...)[0]`, all call sites simplified.

Not addressed (both were flagged as minor):
- `allowedPrefixes: [" "]` — still narrow, still fine to defer.
- `/note` label using the internal `note_id:` name — since option 1 (prefix detection) was chosen, this is still a latent nit but not load-bearing.

**Verdict**: All substantive concerns resolved. Approving.

nvasilevv:
## Post-deploy smoke test

All in-browser, no backend touch needed.

### Happy path (most important)

1. Open a research note in edit mode. On a blank line, type `/` — the popup should appear with 4 items (`/note`, `/filing`, `/chart`, `/embed-chart`).
2. Type `/fil` → list narrows to `/filing`. Press Enter → `[[filing:` is inserted, and the existing `WikilinkAutocomplete` popup opens **pre-filtered to filings with an empty query**. This is the whole fix — if it shows "No results found" with `filing:` as the search term, the hand-off is still broken.
3. Pick a filing. Confirm it renders as a `FilingLinkNode` (styled span with the title).
4. **Save the note, reload the page.** Confirm the filing link is still there — round-trip through markdown works.
5. Repeat for `/note` → note picker, `/chart` → inline `ChartLinkNode`, `/embed-chart` → block `ChartEmbedNode` (chart actually renders, not a link).

### Negative checks (guard against regressions)

6. Type a URL like `https://example.com/path` — slash menu must **not** fire mid-word.
7. Open a chat citation from the feed so the read-only `<TipTapViewer>` renders. Type `/` — popup must **not** appear. This verifies the `readOnly` gate.
8. Open the slash menu, press **Escape** — it dismisses without inserting anything.

### Regression checks for behavior we left untouched

9. In the chat document viewer, open a citation and hit **Ctrl+F** — mark.js highlighting should still work (yellow highlights, Next/Prev navigation).
10. In the notes editor, type `[[` manually (no slash) — the autocomplete should still open as before. Same for `![[` → chart embed picker.

### Sanity

11. Open DevTools Console while using the editor. No red errors from `@tiptap/suggestion`, `tippy.js`, or ProseMirror transaction handling.

If #2 and #4 pass, the feature is working. If #4 fails (link disappears after save/reload), that's a markdown round-trip issue in the custom node serializers, not the slash menu — file separately.

PR #20, opened 2026-04-14
Dedupe issuers by LEI/ISIN/name before creating

nvasilevv:
## Summary
- Add LEI/ISIN (via EODHD) and exact company-name fallback lookups in `get_or_create_issuer` before creating a new row
- Prevents a class of duplicate issuer rows caused by feeds disagreeing on the ticker for the same company

## Related PRs — review & merge order

This PR is part of a three-PR sequence surfaced during the Shelly Group / Digital Networks investigation. They have distinct concerns and should be handled in a specific order.

**Review order:**
1. **#20** (this PR) — the dedup fallback. Introduces `_find_existing_issuer_by_identity`, `_normalize_company_name`, and the overall shape of the logic. #23 leans on this design, so understanding #20 first makes #23 quick to evaluate.
2. **#22** — two-line Qdrant fix (drops a dead `metadata.ticker` filter branch that was timing out the filings search). Fastest review of the three; can be knocked out between deeper passes on #20.
3. **#23** — data cleanup script (`backend/scripts/merge_duplicate_issuers.py`). Lean on the dry-run output in the PR description and the safety-gate logic; the code is straightforward once #20 is internalized.

**Merge / deploy / run order:**
1. **#22** ships whenever — zero interaction with the others, immediately fixes the broken filings search.
2. **#20** merges and deploys to prod next. The new dedup fallback must be live before the migration runs.
3. **#23** merges, then is run manually against prod:
   - dry-run again (confirm cluster counts match the PR description)
   - `--limit 1 --execute` smoke test on the smallest cluster
   - full `--exchange WSE --execute`
   - spot-check 2-3 frontend company pages

The reason #20 must be live before #23 runs: the migration collapses duplicates retroactively, but if the dedup fallback isn't deployed yet, every fresh x3news/gpw filing arriving mid-migration would recreate the duplicates the script just merged. Deploying #20 first closes the intake before you clean up the existing pollution.

## Motivation
Shelly Group SE (LEI `8945007IDGKD0KZ4HD95`, BSE) currently has **three** issuer rows on prod:

| ticker | filings | xbrl_facts | source |
|---|---:|---:|---|
| `SHELLYGROUPS` | 9 | 0 | IR crawler |
| `SLYG` | 42 | 0 | EODHD listings refresh (real BSE ticker) |
| `SELLIGRUPED` | 351 | 4,833 | x3news / KFN XBRL |

All 4,833 XBRL facts — including 1,308 for 2025 reporting periods — live on `SELLIGRUPED`, while the user-facing issuer (`SHELLYGROUPS`) has none. The three rows share the same LEI; two share ISIN `BG1100003166`.

### Root cause
1. `backend/app/services/x3news_adapter.py:38-41` — `derive_ticker` transliterates the Cyrillic company name (`Шелли груп ЕД`) and slugifies it (`SELLIGRUPED`), because the x3news payload contains no ticker, ISIN, or LEI. Every BSE issuer gets a fabricated ticker unrelated to the real exchange symbol.
2. `backend/app/services/repository.py:43-79` — `get_or_create_issuer` previously matched only on `(exchange, ticker)` and `(exchange, eodhd_ticker)`. There was no LEI/ISIN/name fallback, so distinct-ticker feeds produced distinct rows even when they clearly referenced the same issuer.

### Fix
New helper `_find_existing_issuer_by_identity` runs when the ticker/eodhd_ticker lookups miss, in this order:
1. Exact case-insensitive match on `company_name` / `company_name_en`.
2. Legal-form-stripped normalized-name match, using a prefix-narrowed + limited SELECT so the scan stays bounded (first token of the normalized name → `LIKE 'prefix%'`, `LIMIT 200`).
3. ListedSecurity bridge — consult the exchange master-data table by normalized name and use its `issuer_id` or `isin` as a link.
4. EODHD `lookup_identifiers` as the last resort, gated behind all the cheap local lookups. The result is threaded back to the caller so the cold-start creation path can seed the new row without calling `lookup_identifiers` a second time.

Identifier short-circuits: `lookup_identifiers` already no-ops for BSE (`EXCHANGE_MAP` has no entry) and for tickers containing spaces (polluted legal-name values from the gpw poller), so the EODHD branch is zero-cost for the two feeds that generate the worst pollution.

## Scope / follow-ups intentionally not included
- **Data cleanup** of the 42+ existing duplicate clusters → see #23.
- **BSE `ListedSecurity` refresh** — the `listedsecurity` table currently has 0 BSE rows (600 WSE); EODHD doesn't cover the Bulgarian exchange → tracked in issue #21.
- **Feed-side fix** — having the gpw poller resolve the real ticker from `listedsecurity` when `Symbol Emitenta` is empty (would require an async/session-aware parser refactor).

## Test plan
- [x] Unit tests for `_normalize_company_name` — 8 cases covering Polish, Bulgarian, and Western-European legal forms + a suffix-ordering regression guard (`backend/tests/test_company_name_normalizer.py`)
- [x] Unit tests for `_find_existing_issuer_by_identity` — 7 cases covering every branch (exact-name, normalized-name, ListedSecurity-via-id, ListedSecurity-via-isin, EODHD-isin, cold-start, EODHD timeout) using the existing `MagicMock` session pattern (`backend/tests/test_issuer_dedup.py`)
- [ ] Replay an x3news filing for a Shelly Group record in staging and confirm it resolves to an existing issuer rather than creating a new one
- [ ] Verify that the first-ever filing for a new BSE issuer still creates a row (cold-start path)
- [ ] Confirm no regressions for WSE issuers where the ticker match hits on the first lookup


nvasilevv:
## Scope expanded — WSE legal-form pollution

Follow-up finding while investigating Digital Networks SA: **451 of 650 WSE issuer rows (69%)** have a ticker containing a space or dot, i.e. a legal-name abbreviation stored where the ticker should be. Source: the gpw ESPI poller (`backend/app/services/gpw_adapter.py:127-131`), which falls back from `Symbol Emitenta` to `Skrócona nazwa emitenta` (a short legal name like `DIGITAL NETWORK S.A.`) when the symbol field is empty. Bulk audit:

| metric | count |
|---|---:|
| WSE issuers total | 650 |
| ticker contains space or dot (polluted) | 451 |
| clean tickers | 199 |
| duplicate clusters by ISIN | 42 |
| duplicate clusters by LEI | 15 |

The first commit's exact `company_name` match wasn't enough for this class — the GPW payload sends the full legal form (`DIGITAL NETWORK SPÓŁKA AKCYJNA`) while the canonical row has the short form (`Digital Network S.A.`), so exact match misses.

### What this commit adds (004f84e)
- **`_normalize_company_name`** — strips common Polish (SPÓŁKA AKCYJNA ↔ S.A./SA, SP. Z O.O., ASI, SP. K.), Bulgarian (ЕАД/АД/ЕД/ООД/ЕООД), and English (SE/PLC/Ltd/Inc) legal-form suffixes, then collapses whitespace. Both "DIGITAL NETWORK SPÓŁKA AKCYJNA" and "Digital Network S.A." normalize to `digital network`.
- **Normalized-name pass** in `_find_existing_issuer_by_identity` after the exact `company_name` check.
- **ListedSecurity bridge** — for exchanges with populated master data (WSE has 600 rows), looks up `listedsecurity` by normalized name; if it has an `issuer_id` or `isin` set, uses that to locate the canonical Issuer row. This catches cases where the stored Issuer name is too different to normalize but `listedsecurity` knows both.
- Unit tests for the normalizer (`backend/tests/test_company_name_normalizer.py`).

### Still not fixed (intentionally)
- **Cleanup of existing 451 polluted rows + 42+ duplicate clusters** on WSE. Needs a dedicated migration script that merges by LEI/ISIN and rewrites `filing`/`document`/`analystestimate`/etc. FKs. Separable from this PR.
- **gpw_adapter stops emitting polluted tickers at source.** Would want to resolve the real ticker from `listedsecurity` when `Symbol Emitenta` is empty, but that requires making the parser async/session-aware. Out of scope here; the dedup fallback catches it for now.

nvasilevv:
## Summary
This PR adds LEI/ISIN/name fallbacks to `get_or_create_issuer` to prevent the duplicate-issuer rows the Shelly Group case demonstrates. The direction is correct and the motivation is well-documented, but the implementation adds per-ingest external HTTP calls and O(n) table scans to a hot path, and the new dedup helper itself is untested.

## Review Findings

### 🔴 Critical / Security
- **EODHD call on every cache miss in an ingest hot path** (`repository.py:88-94`). `get_or_create_issuer` runs once per inbound filing; any miss now fires an external HTTP request with a 15s timeout. For BSE/x3news the ticker passed in is a synthetic Cyrillic slug (`SELLIGRUPED`) that EODHD will **never** resolve, so every Shelly filing will pay the full round trip before falling through to the name branch. At minimum: (1) negative-cache miss results (e.g. in-process TTL cache keyed on `(exchange, ticker)`), (2) short-circuit for obviously-synthetic tickers, and (3) skip the EODHD call entirely when `company_name` is present and the name branch can answer locally. As written, this can easily regress ingest latency and burn EODHD quota on lookups that are structurally doomed.
- **O(n) table scan inside the name fallback** (`repository.py:140-146` and `165-168`). `SELECT * FROM issuer WHERE exchange = :exchange` loads every issuer for the exchange into Python on every miss, then normalizes each candidate name in a Python loop. Same pattern again for `ListedSecurity`. WSE already has hundreds of rows and this will grow. Either: push the comparison into SQL via a persisted normalized column (`company_name_norm`) with an index, or at minimum add a `LIMIT` + targeted prefix filter. Doing this twice per miss (issuer + listedsecurity) compounds the problem.
- **Race window widened without touching the existing `IntegrityError` retry path.** `get_or_create_issuer` historically relies on a DB uniqueness constraint to catch concurrent creators. The new fallback adds several extra reads before the insert attempt, enlarging the window where two workers both decide "no existing row" and race to create. Please confirm the retry path at the bottom of `get_or_create_issuer` still covers this, and that `Issuer` has the unique constraints you're implicitly trusting. If it doesn't, two x3news workers processing the same filing burst could still produce duplicates despite this PR.

### 🟡 Improvement
- **Dedup helper has zero tests.** `test_company_name_normalizer.py` only covers the pure string function. The interesting logic — EODHD path, ISIN branch, LEI branch, exact-name branch, normalized-name branch, ListedSecurity bridge — is untested. This is write-path code that decides whether to create or reuse a row; please add at least a couple of in-memory `AsyncSession` tests with fixture issuers covering: hit on LEI, hit on ISIN, hit on normalized name, ListedSecurity bridge, and the cold-start miss.
- **Missed opportunity to populate identifiers on the cold-start path.** If `lookup_identifiers` returns an ISIN/LEI but no existing issuer matches, the helper returns `None` and the caller creates a new row *without* those identifiers. Threading the EODHD result back to the caller so a newly-created row is seeded with `isin`/`lei` would prevent the second/third feed from creating another dupe the next time around — which is exactly the failure mode this PR is trying to fix.
- **Legal-suffix list is Europe-incomplete.** Missing common forms you'll hit on GPW/XETRA/similar: `AG`, `NV`, `BV`, `GmbH`, `KGaA`, `SAS`, `Oyj`, `S.p.A.`, `S.r.l.`. Worth adding now since the comment in the source already frames this as the canonical list.
- **Suffix iteration order is load-bearing and brittle.** `_LEGAL_SUFFIXES` mixes very short tokens (`sa`, `ad`, `se`) with multi-word ones. The inner loop `break`s on the first match per pass, so ordering within the tuple silently decides which rule wins. A comment explaining "longest-first / multi-word-first" plus an assertion in a test would protect against reorderings.
- **`logger.info` on every successful dedup** (`repository.py:105-107`, `115-117`, `130-133`, `154-157`, `175-178`, `189-192`). Once this is routine it becomes noise. Drop to `debug`, or keep `info` only for the first-seen-per-exchange case.
- **Deferred imports inside the function.** `from app.services.eodhd import lookup_identifiers` and `from app.models.listed_security import ListedSecurity` are inlined. If this is to avoid a cycle, add a one-line comment; otherwise hoist to the top.

### 🔵 Nitpick
- `_normalize_company_name`'s `while changed` loop is fine but could be a single pass if suffixes are sorted longest-first.
- `try/except Exception` around `lookup_identifiers` is very broad — narrow to `httpx.HTTPError` / `asyncio.TimeoutError` if possible so you don't accidentally swallow programming errors.
- Consider renaming `_find_existing_issuer_by_identity` → `_find_issuer_by_identifier_or_name` since the function's two branches are conceptually different (verified identifier vs. fuzzy name).

## Positive Feedback
- The PR description is excellent: it names the concrete bad rows, traces both root causes (`derive_ticker` slug generation + `get_or_create_issuer` matching keys), and explicitly scopes out the data-cleanup and feed-side fixes. Future readers bisecting this will be grateful.
- Fallback ordering is sensible: verified identifiers (ISIN, LEI) before fuzzy name matching, and exact name before normalized name. That's the right priority.
- `_normalize_company_name` correctly handles the Cyrillic `ЕД` case driving the Shelly bug, and the tests pin down the key GPW/BSE shapes.
- Stripping legal suffixes rather than hand-maintaining an alias table is the right call for v1.

## Conclusion
**Request Changes.** The dedup logic is correct, but it cannot land in the hot path as-is: the per-miss EODHD HTTP call and the full-table issuer scan on a synthetic-ticker feed that *will always miss* are regressions waiting to happen on BSE ingest. Either (a) gate the EODHD + scan behind a cache / early exit so the common BSE path doesn't pay for a lookup that's structurally doomed, or (b) move the dedup pass out of the inline ingest call and into a post-ingest reconciliation job. Also: please add tests for `_find_existing_issuer_by_identity` itself, not just the normalizer, and seed newly-created rows with any identifiers the EODHD lookup returned so the first feed to arrive becomes the anchor for the next one.

nvasilevv:
Thanks for the thorough review. Addressed in 3a513b2 — summary per finding:

## Critical

### EODHD call on every cache miss
**Reordered.** Name-based branches now run *before* EODHD: exact → normalized (prefix-narrowed) → ListedSecurity bridge → EODHD. For the Shelly/x3news path we're already safe because `lookup_identifiers` short-circuits on `EXCHANGE_MAP.get("BSE")` returning `None` before any HTTP call — BSE never paid the round trip. For WSE the reorder means EODHD only runs when every local fallback has missed, i.e. a genuinely new company.

Also threaded the EODHD result back via tuple return: when we *do* call it and no existing row matches, the caller's cold-start seeding block reuses the cached result instead of calling `lookup_identifiers` a second time. That deletes the duplicate EODHD call your "missed opportunity to populate identifiers" note flagged.

Negative caching: deliberately not added. The short-circuits in `eodhd.py:lookup_identifiers` already make BSE/space-containing tickers zero-cost, and for WSE real tickers we *want* the positive result for seeding. I can layer an LRU cache on top if you'd like — LMK.

### O(n) table scan in the name fallback
**Fixed.** The normalized-name path now runs:

```sql
SELECT * FROM issuer
 WHERE exchange = :exchange
   AND (lower(company_name) LIKE :prefix OR lower(company_name_en) LIKE :prefix)
 LIMIT 200
```

The `:prefix` is the first whitespace-delimited token of the normalized incoming name (`digital network spółka akcyjna` → `digital%`). Same treatment for the ListedSecurity bridge query. Python-side normalized compare still runs on the bounded candidate list for correctness, but fan-in is now O(rows-starting-with-prefix) rather than O(exchange-size). No schema change needed.

### Race window
Verified. The unique constraint `uq_issuer_exchange_ticker` on `issuer(exchange, ticker)` still backs the insert path, and I narrowed the `except Exception` on the insert to `except IntegrityError` to make the intent explicit. Two workers processing the same filing for the same (exchange, ticker) still race on the same key and one loses at commit, then falls through to the re-SELECT. Two workers processing two *different* tickers for the same company (the SELLIGRUPED-vs-SHELLYGROUPS class) are now correctly deduped by the name-based fallback — if neither finds an existing row, whichever commits first becomes the anchor and the other's fallback will catch it once it re-enters get_or_create_issuer. The window is slightly wider by a few local SELECTs than before this PR, but the invariant holds.

## Improvements

### Dedup helper zero tests
**Added `backend/tests/test_issuer_dedup.py`** with 7 cases:
1. Exact name match short-circuits (no normalized scan, no ListedSecurity, no EODHD).
2. Normalized-name match via prefix scan — DIGITAL NETWORK legal-form cross-variant.
3. ListedSecurity bridge via `issuer_id` link.
4. ListedSecurity bridge via ISIN.
5. EODHD branch runs last and matches by ISIN.
6. Cold-start returns `(None, identifiers)` so caller can seed.
7. EODHD `httpx.ReadTimeout` is caught without raising.

Follows the existing `MagicMock`-session pattern from `test_notes_service_concurrency.py`. Running with the api container's pytest: 15 tests pass (7 new + 8 normalizer).

### Missed opportunity to seed from EODHD
**Fixed.** Helper returns `(Issuer | None, eodhd_identifiers | None)`. `get_or_create_issuer` now captures the identifiers from the helper and uses them to seed the new row, only calling `lookup_identifiers` itself if the helper never needed to. The post-creation `OpenFIGI` enrichment, URL discovery, and logo lookup are untouched — still run on cold-start as before.

### Legal-suffix list Europe-incomplete
**Added:** `AG`, `A.G.`, `NV`, `N.V.`, `N. V.`, `BV`, `B.V.`, `B. V.`, `GmbH`, `KGaA`, `SAS`, `Oyj`, `S.p.A.` + `SPA`, `S.r.l.` + `SRL`, `S.r.o.` + `SRO`. Cross-variant normalization tests added (Volkswagen AG, ASML N.V. ↔ NV, Enel S.p.A. ↔ SPA, Henkel KGaA, Nokia Oyj).

### Suffix iteration order is load-bearing
**Fixed:** entries are now sorted longest-first in token count, and a module-level `_verify_suffix_order(_LEGAL_SUFFIXES)` runs at import and raises `RuntimeError` if any earlier entry is a trailing-token subsequence of a later one (i.e., if `akcyjna` were placed before `spółka akcyjna`). Pinned by a test: `test_suffix_order_verifier_rejects_bad_ordering`. The normalizer loop is now single-pass now that ordering is deterministic.

### logger.info on every dedup
Dropped all six to `debug`.

### Deferred imports
Hoisted `from app.services.eodhd import lookup_identifiers` and `from app.models.listed_security import ListedSecurity` to module top. No cycles — verified `eodhd.py` and `listed_security.py` don't import from `repository`.

## Nitpicks

- `while changed` loop — replaced with single pass, now that sort order is enforced.
- `try/except Exception` around EODHD — narrowed to `(httpx.HTTPError, asyncio.TimeoutError)`.
- Rename `_find_existing_issuer_by_identity` → didn't change. Happy to rename if you still want, but the fn now covers both identifier *and* name branches, and "by_identity" seemed like a reasonable umbrella. LMK.

## Not changed / follow-ups

- **Out-of-band reconciliation job.** Opted for the in-line fallback because (a) the name branches are now local and cheap, (b) the EODHD branch is zero-cost for BSE and only fires for genuinely new WSE companies, and (c) an out-of-band job leaves a window where duplicates exist before reconciliation. Let me know if you disagree — I can factor the helper into a standalone `reconcile_issuers` job.
- **Existing 451 polluted rows + 42 duplicate clusters.** Still in #23 (`chore/merge-duplicate-wse-issuers`), which must run after #20 deploys.
- **gpw_adapter stops emitting polluted tickers at source.** Still out of scope here — would require making the parser async/session-aware to consult `listedsecurity` during parse.

Test run on the api container:
```
15 passed in 0.58s
```

nvasilevv:
## Follow-up after reviewing 3a513b2

Most of the earlier findings landed cleanly — EODHD reorder, prefix-narrowed queries, cold-start seeding from cached identifiers, expanded suffix list with the longest-first guard, dedup-helper tests, and the polish items (debug-level logs, narrowed excepts, hoisted imports). Verified against the code, not just the response.

Two things I'm still concerned about, plus a few smaller flags.

### 🔴 Prefix-narrowing introduced a new miss class

`_prefix_token` takes the first whitespace-delimited token of the *normalized incoming name* and uses it as a `LIKE 'prefix%'` filter. That silently drops legitimate matches when the incoming feed and the canonical row disagree on the leading noise word:

| incoming (feed) | stored (canonical) | prefix | hit? |
|---|---|---|---|
| `Grupa Benefit Systems S.A.` | `Benefit Systems S.A.` | `grupa%` | ❌ |
| `Grupa Kęty S.A.` | `Kęty S.A.` | `grupa%` | ❌ |
| `Holding X` | `X S.A.` | `holding%` | ❌ |
| `The Digital Network` | `Digital Network S.A.` | `the%` | ❌ |

Pre-PR, the unbounded scan would have caught these via the full normalized compare. Post-PR, the prefix filter discards them before the Python-side normalizer ever sees them. `Grupa` in particular is a very common GPW pattern — the same class of feed inconsistency this PR is trying to solve.

Two reasonable fixes:
1. Strip a small set of leading noise tokens (`grupa`, `group`, `holding`, definite articles) inside `_prefix_token` before picking the first token.
2. Union prefix lookups for every token in the normalized name (still capped at `_NORMALIZED_NAME_CANDIDATE_LIMIT`), so the match can anchor on any meaningful token. Slightly more SQL, still bounded.

I'd take (2) — it's more robust to whichever noise word shows up next. Either way, worth pinning with a regression test that mirrors the `Grupa Benefit Systems` shape.

### 🟡 Race-condition claim is overstated

The response says:

> two workers processing two *different* tickers for the same company (the SELLIGRUPED-vs-SHELLYGROUPS class) are now correctly deduped by the name-based fallback — if neither finds an existing row, whichever commits first becomes the anchor and the other's fallback will catch it once it re-enters `get_or_create_issuer`

That's not what the code does. Trace two concurrent workers with the same `company_name` but different `ticker`:

1. Both miss the exact `(exchange, ticker)` lookup.
2. Both enter `_find_existing_issuer_by_identity`.
3. Both miss the exact-name branch (no row yet).
4. Both miss the normalized-name branch.
5. Both miss the ListedSecurity branch.
6. Both call EODHD and may get the same ISIN/LEI.
7. Both miss the `Issuer.isin` / `Issuer.lei` lookup.
8. Both return `(None, identifiers)`.
9. Both `INSERT` into `issuer` with *different* `ticker` values.
10. `uq_issuer_exchange_ticker` does not conflict because the tickers differ. **Both commits succeed.**

The losing worker never re-enters the helper, because it never lost. Result: two issuer rows for the same company — exactly the bug this PR targets, just in a narrower timing window.

This is acceptable for the scope of this PR because the feed cadence makes simultaneous arrival rare, but the review comment should be accurate. Two options, neither blocking here:

- A partial unique index on `(exchange, isin)` / `(exchange, lei)` when non-null, which would make the losing worker trip `IntegrityError` on commit and fall through to the re-SELECT path (which would then need a widened re-SELECT that also looks up by isin/lei, not just ticker).
- A `SELECT ... FOR UPDATE` sentinel keyed on `(exchange, isin)` / `(exchange, lei)` before insert.

I'd suggest either filing a follow-up issue or editing the earlier reply to explicitly say "same-ticker race is covered; different-ticker concurrent creation remains possible and is out of scope for this PR."

### 🔵 Smaller flags

- **Normalizer collision.** `_normalize_company_name("Airbus SE")` and `_normalize_company_name("Airbus")` both collapse to `"airbus"`. If a holding and a subsidiary sharing a stem name ever coexist on the same exchange, they'll dedup into each other. Edge case, but worth a one-line note in the docstring so the next person isn't surprised.
- **`_verify_suffix_order` is narrower than the comment implies.** It only fires when a shorter suffix is a literal trailing-token subsequence of a longer one in the list — i.e. it would catch `"akcyjna"` before `"spółka akcyjna"`, but not `"sa"` before `"spółka akcyjna"` (because the latter doesn't end in `" sa"`). The current list is safe either way; just soften the comment so it doesn't promise more than it enforces.
- **Still-missing legal forms.** Nordic: `ASA` (Norwegian), `AB` (Swedish), `Oy` (distinct from `Oyj`, Finnish), `A/S` (Danish). Not relevant to PL/BG feeds today, but the source comment frames `_LEGAL_SUFFIXES` as canonical — either add them now or scope the comment explicitly to PL/BG/EN.
- **Tests don't exercise SQL shape.** `_session([...])` scripts exec return values but never validates the query. If a column swap ever lands (e.g. `company_name` → `company_name_en` typo in the WHERE), all 7 tests still pass. Not blocking — consistent with other tests in the repo — but a follow-up integration test against the test Postgres would catch that class of regression.

### Bottom line

Lifting to **Comment** from Request Changes. The two items above are real but the first is a small additive fix and the second is a known-scope acknowledgment. The rest of the PR is in good shape and the core regression — same-company / same-exchange / different-ticker dedup — is now correctly handled for the common case. Happy to ship once the `Grupa` / leading-noise class is addressed (or explicitly tracked as a follow-up).

nvasilevv:
Thanks — all three of the critical/improvement items addressed in f6f5818. Ran the helper trace on paper first this time to catch the gap your Grupa case exposed.

## Prefix-narrowing miss class — fixed

Took **both** (1) and (2), because (2) alone wasn't sufficient. Tracing `Grupa Benefit Systems S.A.` against canonical `Benefit Systems S.A.`:

- Union-of-tokens SQL (your option 2) fixes the candidate **fetch**: `lower(company_name) LIKE 'benefit%' OR LIKE 'systems%'` brings back the canonical row even when `grupa%` would have missed.
- But the Python-side compare after the fetch still normalized to `grupa benefit systems` vs `benefit systems`, so the candidate was fetched then rejected. Writing the test made the gap obvious — it was failing with `StopAsyncIteration` because the helper was falling through to ListedSecurity and then EODHD.
- Added leading-noise stripping inside `_normalize_company_name` itself (your option 1) for: `grupa`, `grupo`, `group`, `holding`, `the`. Loop so `Grupa Holding Foo S.A.` peels both leading tokens. Small curated set — deliberately not open-ended, because aggressive stripping here would collapse unrelated issuers.

Code changes:

```python
def _prefix_tokens(normalized: str) -> list[str]:
    # Every meaningful whitespace-delimited token; len>=2; deduped.
    ...

# in _find_existing_issuer_by_identity:
like_patterns = [f\"{tok}%\" for tok in tokens]
issuer_name_clauses = [
    func.lower(Issuer.company_name).like(p) for p in like_patterns
] + [
    func.lower(Issuer.company_name_en).like(p) for p in like_patterns
]
candidates = (await session.exec(
    select(Issuer)
    .where(Issuer.exchange == exchange, or_(*issuer_name_clauses))
    .limit(_NORMALIZED_NAME_CANDIDATE_LIMIT)
)).all()
```

Same treatment on the ListedSecurity query. Still bounded by `_NORMALIZED_NAME_CANDIDATE_LIMIT = 200`.

Regression tests pinned:
1. `test_leading_noise_word_does_not_block_normalized_match` — drives the exact Grupa/Benefit shape through `_find_existing_issuer_by_identity` with a MagicMock session. Asserts 2 execs (exact miss + per-token candidate list), no ListedSecurity branch, no EODHD.
2. `test_prefix_tokens_ignores_short_tokens_and_dedupes` — pins the token extraction (`len>=2`, dedupe, order).
3. `test_leading_noise_words_stripped` — pins normalizer behavior for `Grupa Benefit Systems`, `Grupa Kęty`, `The Digital Network`, `Grupa Holding Foo` (chained).

## Race-condition claim — acknowledged in code

You're right, my earlier trace was wrong. Walking it again: two workers with the same `company_name` and different `ticker` both reach cold-start, both `INSERT`, `uq_issuer_exchange_ticker` fires on neither (different tickers), both commits succeed. The losing worker never loses.

Rather than backfill the fix into this PR (partial unique index on `(exchange, isin)` + widened re-SELECT is non-trivial and wants its own review cycle), I added it as a **documented known limitation** in the helper docstring so the next reader isn't misled:

```
Known limitations:
- The normalized stem can collapse a subsidiary into its parent
  (e.g. \"Airbus SE\" and \"Airbus\" both normalize to \"airbus\")...
- Same-company / different-ticker concurrent creation is still
  possible: two workers can both reach the cold-start path and
  INSERT different ticker rows before either commits, because
  \`uq_issuer_exchange_ticker\` only catches same-ticker races. A
  partial unique index on \`(exchange, isin)\` / \`(exchange, lei)\`
  would close this; tracked as a follow-up, not in scope here.
```

Will file the follow-up issue after this lands. In practice the feed cadence makes simultaneous arrival for the same issuer rare, but the invariant is worth being explicit about.

## Smaller flags — addressed

- **Normalizer collision** (Airbus SE / Airbus): noted in the docstring limitations block above.
- **`_verify_suffix_order` comment**: softened. Now says \"catches the narrow case where a shorter entry is a trailing-token subsequence of a later, longer one\" and explicitly notes it doesn't enforce a full total order, so new entries still need manual placement.
- **Nordic legal forms**: added `ASA`, `AB`, `Oy` (distinct from `Oyj`), `A/S`. Tests cover each: Equinor ASA, Volvo AB, Kone Oy, Novo Nordisk A/S.
- **Tests don't exercise SQL shape**: agreed, noted but deliberately left for a follow-up — matches existing repo pattern (everything under `backend/tests` uses MagicMock sessions). A Postgres-backed integration test would catch column-swap typos but isn't this PR's scope.

## Test run

```
19 passed in 0.74s
```

(was 15 before: +3 normalizer tests for leading-noise/Nordic, +1 helper test for Grupa, +1 `_prefix_tokens` unit.)

Let me know if you want the partial unique index pulled into this PR instead of a follow-up.

nvasilevv:
LGTM on f6f5818 — verified against the code, not the commit message.

**Prefix-narrowing regression**: fixed both ways I suggested, belt-and-braces. `_prefix_tokens` now ORs a `LIKE` clause per meaningful token (len≥2, deduped, order-preserved) so the match can anchor on `benefit%` or `systems%` rather than only `grupa%`. On top of that, `_normalize_company_name` strips a small curated set of leading noise tokens (`grupa`, `grupo`, `group`, `holding`, `the`) with a loop so chained cases like `Grupa Holding Foo` peel correctly. The `test_leading_noise_word_does_not_block_normalized_match` regression test pins the exact `Grupa Benefit Systems` shape I was worried about, and `test_prefix_tokens_ignores_short_tokens_and_dedupes` pins the tokenizer contract.

**Race condition**: honestly documented in the helper docstring as a known limitation with the `(exchange, isin)` / `(exchange, lei)` partial unique index noted as the follow-up fix. That's exactly what I wanted — either close it or be honest about it, and this is the honest path.

**Smaller flags all picked up**:
- Normalizer stem-collision (`Airbus SE` vs `Airbus`) documented in the docstring.
- `_verify_suffix_order` comment softened to say it catches the narrow trailing-token-subsequence case, not a full total order.
- Nordic forms: `ASA`, `AB`, `Oy`, `A/S` added, with `test_nordic_legal_forms_collapse` covering all four.
- Curated `_LEADING_NOISE_TOKENS` kept deliberately small — the source comment explicitly calls out the collision risk of aggressive stripping, which is the right instinct.

One minor thing worth knowing but not blocking: `AB` and `Oy` are 2-char Nordic suffixes. They only match on a trailing `" ab"` / `" oy"` token, so `Foo Lab` / `Scion Lab` / `Toy` are safe. Swedish companies occasionally use `AB` as a *prefix* (`AB Volvo`), and the leading-noise list doesn't include it — so `AB Volvo` normalizes to `ab volvo` and won't collapse with canonical `Volvo AB` → `volvo`. Extremely narrow edge case, worth knowing if you ever onboard OMX.

**Approve.** Good to ship.

PR #22, opened 2026-04-14
Drop unindexed metadata.ticker branch from Qdrant filter

nvasilevv:
## Summary
- Remove the `metadata.ticker` branch from `_build_filter`'s ticker `should` clause
- Log the swallowed exception in `query_chunks` so silent timeouts are visible

## Root cause
Discovered while debugging a `search_filings(ticker="DIG", filing_date_from="2025-01-01")` call that returned zero results despite DIG having 15 completed, embedded filings in the DB.

`_build_filter` at `backend/app/services/vectorizer.py:311-323` wrapped the ticker match in an OR with three branches:

```python
qdrant_models.Filter(should=[
    Filter(must=[FieldCondition(key="ticker", ...)]),
    Filter(must=[FieldCondition(key="standard_ticker", ...)]),
    Filter(must=[FieldCondition(key="metadata.ticker", ...)]),  # ← unindexed
])
```

The `metadata.ticker` payload key is **not indexed** in Qdrant (confirmed via `GET /collections/ceewire_filings` — it's absent from `payload_schema`). Qdrant has to deserialize the payload of every candidate point to evaluate it, which turns every filtered query into a linear scan.

Measured on prod (777,560 points):

| filter | time | result |
|---|---:|---:|
| `must: [ticker=DIG]` | 0.00s | 3 |
| `should: [ticker=DIG, standard_ticker=DIG]` | 0.03s | 3 |
| `should: [ticker, standard_ticker, metadata.ticker]` (scroll) | **5.04s** | 3 |
| same via `client.search` | **25.77s** | 6 |

The default `qdrant-client` timeout is 5 s. `client.search` raises `httpx.ReadTimeout`, which hits the generic `except Exception: return []` at `vectorizer.py:445-446` and silently returns an empty list. The caller sees "no chunks" and reports "No results found" — with no log line to trace.

## Why the fallback was dead code
100% of points in the collection (777,560 / 777,560) have `ticker` indexed as a top-level keyword. Zero points have been embedded with a nested `metadata.ticker` — the branch was an older defensive path that was never actually load-bearing.

## Fix
- Collapse the ticker `should` to just `ticker` and `standard_ticker`, both indexed keyword fields.
- Add `logger.exception` before the empty-list return in `query_chunks` so this class of bug can't hide behind a silent zero-result again.

## Test plan
- [ ] Re-run `search_filings(ticker="DIG", filing_date_from="2025-01-01")` and confirm non-zero results
- [ ] Spot-check a few sparse-ticker searches on other WSE companies
- [ ] Verify common-ticker searches (which used to succeed) still return sensible top-K results


nvasilevv:
# Review — Drop unindexed `metadata.ticker` branch from Qdrant filter

## Summary
Small, surgical fix (+8/−10) that removes a dead `metadata.ticker` branch from `_build_filter`'s ticker `should` clause and adds a `logger.exception` before the catch-all empty-list return in `query_chunks`. The PR description is exceptionally well-researched: it pinpoints a linear scan caused by an unindexed payload key, backs it with production measurements (777,560 points, 25.77 s vs 0.00 s), and proves the removed branch is dead code (zero points carry `metadata.ticker`).

## Findings

### 🔴 Critical / Security
None. No injection surface, no new deps, no secrets, no authz changes.

### 🟡 Improvement (non-blocking, please fold into this PR)

- **`vectorizer.py:430-438` — stale workaround that this PR obsoletes.** The "retry once if Qdrant returns 0 results with a ticker filter (intermittent Qdrant segment/index issue)" block with a 0.5 s `time.sleep` is almost certainly a band-aid for the very timeout pathology you just diagnosed: the first `client.search` would silently time out, return `[]`, and the retry would occasionally beat the deadline. Now that the root cause is fixed, this retry is dead weight — it doubles latency on every genuinely-empty ticker query (sparse tickers, typos) and adds a sync `sleep` inside a request path. Please delete it as part of this PR.
- **Log payload is incomplete.** `logger.exception("query_chunks failed (ticker=%s, filing_date_from=%s)", ticker, filing_date_from)` omits `sector`, `filing_ids`, and `filing_date_to`. If the next silent-timeout regression is on a sector or date-range filter, this log won't help reproduce it. Please include all five filter args.
- **`vectorizer.py:448-449` — catch-all `except Exception: return []` is still too wide.** You correctly added `logger.exception`, which is the main win, but the behavior of swallowing *every* exception into an empty-result response is exactly the shape that hid this bug. Consider narrowing to `httpx.ReadTimeout` / `qdrant_client.http.exceptions.UnexpectedResponse` and re-raising anything else so programming errors surface loudly. OK as a follow-up if you'd rather not expand scope here.

### 🔵 Nitpick
- The new inline comment is ~5 lines and restates what the PR description already explains. A one-liner like `# metadata.ticker dropped: unindexed, forces full payload scan (see PR #22)` is enough and ages better than the prose version.
- `logger.exception` already attaches the traceback, so the message string can be shorter — the noun is "query_chunks failed," the format args carry the rest.

## Positive feedback
- **Diagnosis quality is the standout.** The write-up isolates the problem to an unindexed payload key, measures it against indexed alternatives, connects the 5 s qdrant-client default to the 5.04 s scroll time, and traces the symptom all the way to `except Exception: return []`. Exactly the bar for deleting a defensive fallback.
- **Pairing the fix with the log line** is the right move: even if something else silently returns `[]` in future, the next engineer gets a traceback instead of a mystery.

## Conclusion
**Approve** once the two non-blockers above (delete the 0.5 s retry, expand the log args) are folded in. The narrower-except cleanup can be a follow-up. Core change is correct, minimal, and well-justified.

nvasilevv:
✅ **Approve**

All three non-blockers from the earlier review are addressed:

- Stale 0.5 s retry + `import time as _time` deleted from `query_chunks`.
- `logger.exception` now includes all five filter args (`ticker`, `sector`, `filing_ids`, `filing_date_from`, `filing_date_to`).
- Inline comment collapsed to the one-liner.

Narrowing the catch-all `except Exception` in `query_chunks` is fine to leave as a follow-up. LGTM — ready to merge.

PR #23, opened 2026-04-14
Add merge_duplicate_issuers maintenance script

nvasilevv:
## Summary
One-shot cleanup script for the duplicate issuer rows surfaced in #20. Groups issuer rows by shared ISIN / LEI on the same exchange, picks a canonical row per cluster, repoints FK references (`filing`, `document`, `analystestimate`, `backfilljob`, `ir_*`, `listedsecurity`, `watchlistitem`, `datasetrow`, `issuernarrative`), copies missing identifiers from losers onto the canonical row, and deletes the losers. Dry-run by default.

## Canonical row selection
Sort each cluster by, in order:
1. **clean ticker** — no space, no dot, ≤ 6 chars (so `DIG` beats `DIGITAL NETWORK S.A.`)
2. **activity** — `filing + document` row count
3. presence of `isin`, `lei`, `figi`
4. older `created_at` as a deterministic tiebreaker

## Safety gate
Clusters are skipped when any of:
- **ISINs conflict** (e.g. `EUROCASH` has `PLEURCH00037` and `PLEURCH00011` — likely an old/new ISIN after a corporate action)
- **LEIs conflict** (several rows have one correct LEI and one with a KRS number concatenated on the end — `259400...0000704863` — which is a separate data-quality bug)
- **LEI-only** (no ISIN on any row in the cluster)

`--force` opts into processing these anyway.

### Why the gate matters — real false positive it caught
```
WINNER  ticker='ZABKA'        isin='PLO437000014'  lei='259400S5N7WSF3HDGM86'
loser   ticker='CPI FIM S.A.' isin='LU0122624777'  lei='259400S5N7WSF3HDGM86'
```

Żabka Polska and CPI FIM S.A. are obviously different companies but share an LEI in our DB because of a bad LEI value on one of the rows. Without the gate, the script would have merged them and destroyed real data.

## Prod dry-run (WSE)
```
found 49 duplicate cluster(s) (exchange=WSE)
41 cluster(s) safe to merge, 8 skipped

=== totals (dry-run) ===
  analystestimate                18
  document                       1878
  filing                          1287
  ir_asset                         6
  ir_crawl_source                  1
  ir_page_snapshot                 1
  issuer:backfill                 153
  issuer:delete                    42
  issuernarrative:delete           38
  issuernarrative:repoint          19
  listedsecurity                   6
  watchlistitem                    3
```

Notable safe clusters:
- `ATT` + `PUŁAWY` + `GRUPA AZOTY SA` — 3-way merge, all share `PLZATRM00012`
- `SCP` ← `SCOPE FLUIDICS S.A.` (96 + 7 filings)
- `KRUK` ← `KRUK S.A.` (152 + 180 filings, both carry meaningful history)
- `DAT` ← `DATAWALK S.A.` (67 + 59 filings)
- `BANK` ← `BANK PEKAO S.A.` (40 + 66 filings)

## Usage
```bash
# dry-run (default) — just logs the plan
python -m scripts.merge_duplicate_issuers --exchange WSE

# dry-run including unsafe clusters
python -m scripts.merge_duplicate_issuers --exchange WSE --force

# actually execute
python -m scripts.merge_duplicate_issuers --exchange WSE --execute
```

## Ordering with PR #20 and #22
This script **must run after** PR #20 (`fix/bse-issuer-dedup`) is deployed to prod. Otherwise, fresh x3news / gpw filings arriving during the migration would recreate the duplicates the script just merged (since the dedup fallback in `get_or_create_issuer` isn't live yet).

PR #22 (Qdrant `metadata.ticker` fix) is independent and can land in any order.

## Follow-ups intentionally not in this PR
- **Skipped unsafe clusters** need a separate manual cleanup pass — mostly LEI+KRS concatenation bugs and one-off bad data. Each one benefits from human judgment more than heuristics.
- **BSE** support works via `--exchange BSE` but is currently empty (the only BSE cluster — Shelly Group — was merged by hand earlier today).
- **Qdrant repointing**: the script updates Postgres but not Qdrant chunk payloads. Their `issuer_id` / `ticker` fields remain stale. For the current filter logic this is fine (search filters on `ticker` which stays valid via backfill), but long term we should add a re-embed / payload-update pass.

## Test plan
- [ ] Re-run dry-run after PR #20 deploys to confirm cluster counts match
- [ ] `--limit 1 --execute` as a smoke test on the smallest cluster first
- [ ] Full `--execute` run, then verify: (a) no remaining ISIN duplicates in the target exchange, (b) a sample canonical issuer has all formerly-loser filings, (c) the Qdrant collection still returns results for affected tickers
- [ ] Spot-check 2-3 frontend company pages for merged issuers


nvasilevv:
## Review — PR #23: `Add merge_duplicate_issuers maintenance script`

### Summary
One-shot maintenance script that clusters duplicate `issuer` rows by shared ISIN/LEI on the same exchange, picks a canonical row, repoints FKs across ~10 tables, backfills identifiers, and deletes losers. Dry-run by default, gated by explicit safety checks. A single 483-line additive file with no production code paths touched.

### Review Findings

#### 🔴 Critical / Security
- **Unique-constraint violations on FK repoint.** `watchlistitem` typically has a `UNIQUE(user_id, issuer_id)` constraint, and `listedsecurity` has uniqueness on `(exchange, ticker)` / `(issuer_id, ticker)` in most schemas I've seen in this repo. If any user has watch-listed both the winner and a loser — or both issuers have overlapping listed securities — the plain `UPDATE ... SET issuer_id = :winner WHERE issuer_id = ANY(:losers)` will hit a constraint error and the whole cluster rolls back. The try/except at `merge_duplicate_issuers.py:402-407` will swallow it and move on, leaving silently un-merged clusters. **Action:** before merging, verify constraints on `watchlistitem` / `listedsecurity`; if they exist, dedupe losers against the winner (delete colliding loser rows) before the `UPDATE`. At minimum, log which cluster failed with the constraint name so the manual cleanup pass can find them.

#### 🟡 Improvement
- **Narrative ordering is correct but fragile.** `_merge_cluster` relies on the order: delete losing narratives → repoint kept narrative → (implicitly) winner's old narrative was in the delete set if it wasn't the "best". This works, but the invariant isn't obvious at a glance. A one-line comment on `merge_duplicate_issuers.py:311` ("deleted first so the UNIQUE(issuer_id) repoint below can't collide") would save the next reader a re-derivation.
- **`_fk_counts` is N×M queries per cluster.** For 49 clusters × 11 tables that's ~540 round-trips — fine for a one-shot, but a single `UNION ALL` over all losers/winners up front would cut it to one query and make dry-run noticeably snappier on bigger exchanges. Not worth doing for this PR, but worth noting if you end up re-running it.
- **`configure_logging()` at module import** (`merge_duplicate_issuers.py:68`) is a side effect on import. If anything ever `from scripts.merge_duplicate_issuers import ...` it reconfigures the root logger. Move it inside `main()`.
- **Dry-run "backfill count" semantics disagree with execute path.** In dry-run, `repoint_counts["issuer:backfill"] = len(backfill)` counts *fields*; in execute it also counts fields (not rows). That's internally consistent, but in the totals log it reads as if 153 issuer rows will be backfilled when actually it's "153 field writes across N issuers." A quick rename to `issuer:backfill_fields` would avoid confusion when reading the prod dry-run output you pasted in the PR body.
- **`narrative_rows` max by `filing_count_at_generation or 0`** at `merge_duplicate_issuers.py:298`: if all narratives are `None`, `max()` picks whichever comes first in query order, which is nondeterministic across Postgres runs. Either add `ORDER BY id` to the query or tiebreak in Python on `row[0]`.

#### 🔵 Nitpick
- `from typing import Iterable` (`merge_duplicate_issuers.py:24`) is unused.
- `for l in losers` uses `l` as an identifier; E741 / readability. `loser` is a two-char win.
- `warnings.filterwarnings(...)` is interleaved with imports (`merge_duplicate_issuers.py:22` sits between two `import` blocks). Lint tools that enforce `__future__` → imports → code ordering will flag this. Move the filter inside `main()` or after all imports.
- The safety-gate docstring says "Only ISIN-consistent clusters auto-merge" but the function also rejects conflicting LEIs and LEI-only clusters. Minor doc drift.
- `_is_clean_ticker` hardcodes `≤ 6 chars` — fine as a heuristic, but a module-level `_CLEAN_TICKER_MAX_LEN = 6` makes the rule greppable if WSE/BSE ever diverge.

### Positive feedback
- **The safety gate is the right call, and the Żabka/CPI anecdote in the PR body is exactly the kind of thing that justifies it** — merging on LEI alone would have silently destroyed real data, and the explicit `--force` opt-in is the correct ergonomic. This is the kind of defensive thinking I want in a destructive one-shot.
- **Snapshotting ORM instances immediately into `IssuerSnapshot` dataclasses** (`merge_duplicate_issuers.py:71`) sidesteps the whole class of "expired attribute after commit/rollback" bugs that async SQLModel makes easy to hit. Clean choice.
- **Union-find across LEI and ISIN groups** handles the case where three rows share an ISIN with one pair and an LEI with another — correct and surprisingly rare to see done right in ad-hoc scripts.
- **Parameterized `text()` queries with hardcoded constant table names** — no injection surface, and the pattern is obvious enough that a future reader won't be tempted to pass user input through.
- **Transaction-per-cluster with rollback-on-exception** means one bad cluster can't poison the whole run. Good blast-radius control.

### Conclusion
**Request Changes** — but only on the 🔴 item. The uniqueness-constraint question on `watchlistitem` and `listedsecurity` is load-bearing: if either has a conflict in prod, the `--execute` run will silently skip clusters that looked safe in dry-run, and you'll find out by reading logs afterward. Confirm the constraints (a quick `\d watchlistitem` / `\d listedsecurity` on prod), and either add pre-dedupe UPDATEs or document that clusters failing on those tables are expected in the manual follow-up list. Everything else is cleanup that can land in a follow-up.

nvasilevv:
## Follow-up review — commit `1383f47f`

Verified the author's claims and walked the diff against my original findings.

**🔴 Critical — resolved.** Author's commit message claim checks out:
- `backend/app/models/watchlist.py:33` — `WatchlistItem` has no `__table_args__` / no `UniqueConstraint`, so repointing `issuer_id` can't violate anything.
- `backend/app/models/listed_security.py:14` — `UniqueConstraint("exchange", "ticker")`, not on `issuer_id`. Repointing doesn't touch the constrained columns.

So neither table can collide on the FK update. The load-bearing concern was real in principle but not in this schema. Good that they also improved the except handler at `merge_duplicate_issuers.py:447-457` to log exception type, ids, and tickers — if a surprise constraint shows up during `--execute`, it'll now be findable instead of a silent rollback.

**🟡 Improvements — all addressed:**
- Narrative delete-before-repoint invariant documented (`:298-301`).
- Deterministic narrative pick: `ORDER BY id` + `(r[2] or 0, str(r[0]))` tiebreak (`:307, :316`).
- `configure_logging()` moved into `main()` (`:491`).
- `issuer:backfill` → `issuer:backfill_fields` in both dry-run and execute paths.
- N×M `_fk_counts` left as-is (I flagged it as not-worth-doing; agreed).

**🔵 Nitpicks — all addressed:** unused `Iterable` gone, `l` → `loser`, `warnings.filterwarnings` moved below imports, docstring now covers LEI-conflict + LEI-only cases, `_CLEAN_TICKER_MAX_LEN` extracted.

**Verdict: LGTM, approve.** Every item from the prior review is either fixed or has a justified non-fix. Nothing new introduced.

PR #24, opened 2026-04-14
Reconcile Qdrant chunk payloads after issuer merges

nvasilevv:
## Summary
- Add `scripts/reconcile_qdrant_issuers.py`: scrolls the collection once, classifies each distinct `issuer_id` as `clean` / `drifted` / `repointed` / `orphan` against current Postgres state, and rewrites payloads. Dry-run by default.
- Wire an inline Qdrant repoint hook into `scripts/merge_duplicate_issuers.py` (#23) so future `--execute` runs don't create new drift.
- Factor out three small helpers in `vectorizer.py` — `count_points_by_issuer`, `reassign_issuer_on_points`, `delete_points_by_issuer` — reused by both scripts.

## Why
`merge_duplicate_issuers.py` from #23 repoints `filing.issuer_id` in Postgres but never touches Qdrant. Chunk payloads keep the *loser* `issuer_id` / `ticker` / `company_name`, so `search_filings(ticker="DIG")` returns stale results whose parent filings are now owned by a different canonical issuer. #23's PR body explicitly flagged this as a follow-up.

Concrete trigger: `search_filings(ticker="DIG")` returned 6 chunks across 3 filings while `query_filings_db` saw only 2 — the third PDF was stamped in Qdrant against a pre-merge loser issuer_id.

## How reconcile classifies drift
For each distinct `issuer_id` currently stamped on Qdrant points, `classify_sample` returns one of:

| action | meaning | fix |
|---|---|---|
| `clean` | issuer exists in Postgres, sample payload matches | no-op |
| `rewrite` | same issuer_id, stale ticker / company_name / figi / … | `set_payload` keeping the same issuer_id |
| `repoint` | issuer deleted, but sample `filing_id` is still in Postgres under a live issuer (merge-script case) | `set_payload` onto the live issuer |
| `orphan` | issuer deleted, sample filing also gone | logged; `--delete-orphans` to prune |

Scan cost on prod (~770k points): one payload-only scroll, ~400 round trips at `limit=2_000`, pulling only the 7 fields we compare.

## Merge-script hook
`_merge_cluster` now calls `vectorizer.reassign_issuer_on_points(loser_id → winner_payload)` after the Postgres `DELETE FROM issuer`. The call is wrapped in try/except so a Qdrant outage can't undo a commit that's already about to happen — the standalone reconcile script remains the backstop. `--skip-qdrant` on the merge script opts out entirely for environments where Qdrant isn't reachable.

The winner payload is computed from the snapshot **plus** any fields the winner is about to backfill from losers (`_qdrant_payload_for_winner`), so Qdrant and Postgres end up with the same final values.

## Test plan
- [x] `tests/test_vectorizer_reassign.py` — 4 tests mocking `QdrantClient` (happy path, no-op on empty filter, failure returns 0, `delete_points_by_issuer` uses `FilterSelector`)
- [x] `tests/test_reconcile_qdrant_issuers.py` — 6 tests on the pure `classify_sample` function covering every branch, including `None` → `""` coercion and empty `filing_id` bridging
- [x] Local run: 30/30 (new 10 + regressions for `test_vectorizer.py`, `test_issuer_dedup.py`, `test_company_name_normalizer.py`)
- [ ] Dry-run `python -m scripts.reconcile_qdrant_issuers` against prod, eyeball cluster counts for `drifted` / `repointed` / `orphaned`
- [ ] If counts look sane: `--execute`, then re-run the original `search_filings(ticker="DIG")` vs `query_filings_db` comparison and confirm they agree
- [ ] Spot-check Shelly Group (`SELLIGRUPED` → `SLYG`) in the frontend after the execute run

## Ordering with prior PRs
- #20 (dedup fallback) → already merged
- #22 (Qdrant `metadata.ticker` filter fix) → already merged
- #23 (merge script) → already merged, but any cluster merged before **this** PR ships will have left Qdrant drift. The reconcile script is specifically designed to clean that up in one pass.


nvasilevv:
# PR Review — Reconcile Qdrant chunk payloads after issuer merges

## Summary
Adds a standalone reconcile script that scans Qdrant, classifies each distinct `issuer_id` against Postgres as clean/drifted/repointed/orphan, and rewrites payloads — plus wires an inline Qdrant repoint hook into the existing `merge_duplicate_issuers.py` script so future merges don't create new drift. Three reusable helpers (`count_points_by_issuer`, `reassign_issuer_on_points`, `delete_points_by_issuer`) are factored into `vectorizer.py`. Well-scoped follow-up to #23 with tests covering both the helpers and the pure classifier function.

## Review Findings

### 🔴 Critical / Security
| # | Finding |
|---|---|
| C1 | **`id = ANY(:ids)` with string UUIDs may fail under asyncpg.** `reconcile_qdrant_issuers.py:_load_issuers` and `_load_filing_issuer_map` pass `list[str]` into `text("SELECT ... WHERE id = ANY(:ids)")`. Postgres `issuer.id` / `filing.id` are UUID columns; asyncpg does **not** implicitly cast `text[] → uuid[]`, and the raw SQLAlchemy `text()` path bypasses SQLModel's UUID coercion. Expect `DataError: invalid input for query argument $1: expected uuid`. Fix: either cast server-side (`WHERE id = ANY(CAST(:ids AS uuid[]))`) or convert to `UUID` in Python before passing. Worth actually running the dry-run once before merge — the tests exercise only `classify_sample`, so this path is completely untested. |

### 🟡 Improvement
| # | Finding |
|---|---|
| I1 | **Inconsistency window between Qdrant rewrite and Postgres commit in merge script.** `merge_duplicate_issuers.py:_merge_cluster` calls `reassign_issuer_on_points` **inside** the transaction, after the `DELETE FROM issuer` but before the caller commits. If anything fails between `_merge_cluster` returning and `session.commit()`, Qdrant is permanently rewritten while Postgres rolls back the entire cluster. The backstop is the reconcile script, but that's a silent data-integrity window worth a comment at minimum — or moving the Qdrant call to *after* `session.commit()` in `merge_duplicates` (the caller), so Postgres is the source of truth and Qdrant follows. |
| I2 | **`_qdrant_payload_for_winner` ignores `backfill` for `ticker` and `company_name`.** The helper reads `winner.ticker` / `winner.company_name` directly instead of going through `_final()`. Today the merge script never backfills these fields (they drive winner selection), but the asymmetry is a latent bug — if a future merge ever backfills one, Qdrant will silently drift. Cheap to make symmetric. |
| I3 | **`reassign_issuer_on_points` issues a `count` before every `set_payload`.** In a bulk reconcile over thousands of drifted issuers that's 2× the round trips. The count is only used to (a) short-circuit no-ops and (b) return a number for logging. `set_payload` on a filter is already a no-op when the filter matches nothing; consider dropping the count in the reconcile path and relying on the post-update log line. Keep it for the merge-script path where you want the "0 points touched" signal. |
| I4 | **Two separate `get_session_context()` blocks in `reconcile()`** — one for the initial load and another only if `bridge_ids` is non-empty. Merge into one session, or load bridges in the first query (`issuer.id IN (sample_ids ∪ (SELECT issuer_id FROM filing WHERE id = ANY(missing_filing_ids)))`). Minor. |
| I5 | **Return value ambiguity on failure.** `reassign_issuer_on_points` returns `0` for both "no points matched" and "set_payload raised." Callers can't distinguish a clean no-op from a swallowed Qdrant outage. For the merge-script hook this matters — you don't know whether to recommend running reconcile afterwards. Consider returning `-1` / raising a typed exception on real failures. |
| I6 | **Scroll offset semantics.** `_scan_qdrant` breaks both on empty `points` and on `next_offset is None`. The `if not points: break` before the for-loop is a dead safety check (qdrant_client returns a non-empty page whenever `next_offset` is not None); leave it, but note the loop would also terminate forever if `next_offset` stays stable and `points` is non-empty — add an upper bound on iterations as a defensive cap on a 770k-point collection. |

### 🔵 Nitpick
- `reconcile_qdrant_issuers.py` uses `dataclasses.dataclass` for `PointSample` / `IssuerRow` — fine, but the file also imports `dataclasses` just for `@dataclasses.dataclass`; `from dataclasses import dataclass` is more idiomatic and avoids the `dataclasses.dataclass` namespace at each use site.
- `logger.info("  would %-8s %s → %s (%s)", ...)` — the leading two-space indent inside log format strings is unusual. Consider a dedicated plan-printing helper rather than visual indentation in logs.
- The assertion `assert new_payload is not None` in `reconcile()` after `action == "rewrite"` is load-bearing only under `-O`-off; a narrower type (`tuple[Literal["clean"], None] | tuple[Literal["rewrite","repoint"], dict]`) would let mypy prove it.
- `test_reconcile_qdrant_issuers.py:test_missing_issuer_with_empty_filing_id_is_orphan` asserts `orphan` but the comment explains WINNER exists — add a second assertion that `_` payload is `None` for completeness.

## Positive Feedback
- **`classify_sample` as a pure function is excellent** — no DB, no Qdrant, just data in → decision out. That's why the test file can import it without dragging in `get_session_context`, and it's the right shape for a classifier. Lazy imports inside `reconcile()` to preserve that purity is a nice touch.
- **Coercing `None → ""` at the vectorizer boundary** matches the upsert contract and explicitly protects the keyword indexes — good to see this documented in the docstring and pinned by `test_reassign_issuer_on_points_rewrites_payload_fields`.
- **Bridge-via-`filing_id` classification** is the clever bit. It turns "loser issuer vanished from Postgres" from an unrecoverable orphan into a repointable case, which directly addresses the merge-script aftermath. Worth the explicit `repoint` action in the enum rather than collapsing it into `rewrite`.
- **`--skip-qdrant` flag on the merge script** is the right escape hatch for Qdrant-unreachable environments, with the reconcile script as the documented backstop. Good separation of concerns.
- **Test coverage of the vectorizer helpers** (`FilterSelector` check, `set_payload` failure returns 0, None-coercion) pins the exact Qdrant API shape future refactors might accidentally break.

## Conclusion
**Request Changes** — primarily for **C1** (UUID/asyncpg type mismatch in the raw-SQL path will likely break the dry-run on its first real invocation, and it's uncovered by tests). **I1** (Qdrant-before-commit ordering in the merge hook) is worth fixing or at minimum documenting before this lands, since it widens the exact integrity gap #23 was flagged for. Everything else is polish. Once C1 is fixed and you've actually dry-run this against a real Postgres + Qdrant, this is a clean merge.

nvasilevv:
# Follow-up review — revision `2a2dcdc`

Revision addresses all six points from the previous pass cleanly.

## ✅ Fixes verified

**C1 (UUID/asyncpg)** — `_load_issuers` and `_load_filing_issuer_map` normalize `list[str] → list[UUID]` before handing to `text()`. Inline comment explains why.

**I1 (Qdrant-before-commit ordering)** — The fix is exactly the right shape, and better than what I originally suggested. Instead of just reordering calls, `_merge_cluster` now returns a `ClusterResult(counts, qdrant_plan)` dataclass. The plan captures `winner_payload` + `loser_ids` at the moment they're still valid; the caller runs `_apply_qdrant_plan(result)` **only after** `session.commit()` succeeds. Postgres is unambiguously the source of truth, and a Qdrant outage can't undo a committed merge. `_apply_qdrant_plan` correctly swallows `QdrantReassignError` (the commit already happened — rolling back isn't an option) while logging the "run reconcile afterwards" hint.

**I2 (symmetry)** — `ticker` and `company_name` go through `_final()` now. Docstring updated to explain why.

**I3 (skip the count round trip)** — `reassign_issuer_on_points` no longer pre-counts; signature changed to `-> None`. The merge script gets its tally by calling `count_points_by_issuer` explicitly in `_apply_qdrant_plan` (it wants the number in the summary). Reconcile skips the count entirely and tallies at the issuer level instead of the point level — the log line was updated to say `issuers_rewritten` rather than `points_rewritten`, which is honest.

**I4 (two sessions)** — Bridge fetch moved inside the first `async with get_session_context()` block.

**I5 (error ambiguity)** — New `QdrantReassignError` exception. Reconcile catches it per-plan-item and increments a `failures` counter surfaced in the totals log. Merge-script path catches and logs (right call post-commit). Test renamed to `test_reassign_issuer_on_points_raises_on_set_payload_failure` and uses `pytest.raises`.

**I6 (scroll termination)** — `_SCROLL_MAX_ITERATIONS = 10_000`, `for…else` termination clause that logs an error if the cap is hit. Comment notes 20M points at batch=2000, two orders of magnitude above current collection size.

**Nits** — `from dataclasses import dataclass` done; orphan test now asserts `payload is None`; new `test_count_points_by_issuer_returns_count` added; `test_reassign_issuer_on_points_rewrites_payload_fields` now asserts `client.count.assert_not_called()` to pin I3 at the test level — nice touch, prevents a regression of "helpfully adding back the pre-count."

## 🔵 New minor observations (non-blocking)

- **TOCTOU inside `_apply_qdrant_plan`** — it calls `count_points_by_issuer` then `reassign_issuer_on_points`, and if a point is upserted against the loser in between, the tally is off by a few. The `set_payload` still does the right thing on the filter, so this is just the operator's reported count being slightly stale, not a correctness issue. Fine as-is.
- **`assert result.qdrant_plan is not None`** in `_apply_qdrant_plan` relies on `-O` being off and on the caller checking `result.qdrant_plan is not None` first. The caller in `merge_duplicates` does check it (`if not skip_qdrant and result.qdrant_plan is not None`), so the invariant holds. A `None`-guarded early return at the top of `_apply_qdrant_plan` would be a shade safer than the assert, but genuinely a nit.

## Conclusion
**Approve.** The I1 refactor in particular is the kind of response that's clearly better than the original, not a minimal patch to satisfy a reviewer — moving to a plan-then-apply split makes the ordering constraint structural rather than comment-enforced. Dry-run against prod still worth doing before the `--execute`, but nothing in the code itself is blocking.

nvasilevv:
# Follow-up review — revision `b3e7aeb`

Targeted fix for a real prod failure — `--execute` hit `ResponseHandlingException: timed out` because `issuer_id` wasn't in `_PAYLOAD_INDEX_FIELDS`, so every filtered `set_payload` fell back to a full payload scan on ~800k points and blew past qdrant-client's 5s default. Same failure mode #22 fixed for `metadata.ticker`. Verified end-to-end on prod: 70/70 issuers repaired, follow-up dry-run reports 802 issuers 100% clean, DIG acceptance test green.

## ✅ The fix itself
- **`issuer_id` added to `_PAYLOAD_INDEX_FIELDS`** — correct. `_ensure_payload_indexes` runs on every `ensure_collection()` call even for existing collections and back-fills missing indexes, so the next upsert will self-heal old collections too.
- **`_ensure_issuer_id_index` in reconcile** — pragmatic self-heal. Uses `get_collection().payload_schema` to check, creates with `wait=True` so subsequent `set_payload` calls use the index immediately.
- **`timeout=600`** — 10 minutes covers both the index build and cumulative `set_payload` round trips.
- **Warning filter** — mirrors the pattern already in `merge_duplicate_issuers.py`.

## 🟡 Concerns (non-blocking but worth flagging)

1. **Merge-script hook path still vulnerable on first post-deploy run against an old collection.** `_apply_qdrant_plan` uses `vectorizer._qdrant()` which constructs a client with qdrant-client's **default 5 s timeout** — no `timeout=600`. If someone runs `merge_duplicate_issuers.py --execute` on a collection where the `issuer_id` index hasn't been back-filled yet (i.e., no upsert has run since deploy, and reconcile hasn't been run either), the hook will hit the exact same full-scan timeout that reconcile just fixed. The mitigation that exists — `_ensure_payload_indexes` back-filling during upsert — only helps *if* an upsert happens first. Options:
   - Add `_ensure_issuer_id_index` (or a general `_ensure_payload_indexes` call) at the top of `_apply_qdrant_plan` too.
   - Bump `_qdrant()`'s default timeout — affects all vectorizer callers though, probably too broad.
   - Document in the merge-script docstring that reconcile should be run first on old collections.

   Prod is already repaired out-of-band, so this isn't urgent — but any environment that hasn't been through that one-time cleanup is exposed.

2. **`_ensure_payload_indexes` creates indexes with `wait=False`.** Unchanged by this PR, but now relevant: on a ~800k-point collection, the `issuer_id` index build takes real wall time. If `ensure_collection` triggers the back-fill and `upsert_chunks_with_vectors` follows within seconds, the upsert's own filter operations (if any) race against index readiness. Not a new issue — existing fields have the same property — but worth being aware of.

3. **`_ensure_issuer_id_index` is not guarded with try/except.** `_ensure_payload_indexes` in vectorizer swallows failures as "non-critical" so upserts keep working. The reconcile version will propagate any `get_collection` or `create_payload_index` exception up and abort the whole reconcile run. That's probably the right call for reconcile specifically (you want to know before you start rewriting 100+ payloads), but worth noting as an intentional asymmetry.

4. **Reconcile now opens its own `QdrantClient(url=..., timeout=600)` instead of going through `vectorizer._qdrant()`.** Fine for now — reconcile legitimately wants a longer timeout — but it means future changes to `_qdrant()` (auth headers, TLS, pooling) won't be picked up. If `_qdrant()` ever grows configuration beyond the URL, reconcile will silently drift. Minor, filed for future awareness.

## Verdict
**No blocking issues.** Concern #1 is the only one I'd genuinely want addressed, and even that only matters if someone runs the merge script before an upsert or reconcile has touched an old collection. Given prod is already repaired, it's a latent footgun rather than an active bug. A one-liner adding `_ensure_issuer_id_index(vectorizer._qdrant())` at the top of `_apply_qdrant_plan` — or just a comment in the merge-script docstring — would close it.

nvasilevv:
# Follow-up review — revision `8850a5c`

Clean fix for concern #1 from the previous pass. Shared `vectorizer.ensure_issuer_id_index(client=None)` helper, optional client with self-owned `timeout=600` fallback and proper `try/finally` close. Reconcile now delegates to the shared helper with its own long-timeout client (no second client opened just for the index check). Merge script calls it once per process via a `_qdrant_indexes_ensured` module-level flag at the top of `_apply_qdrant_plan`, wrapped in try/except so a failure just logs the "run reconcile afterwards" hint and lets the merge loop continue — correct, since the Postgres commit has already happened.

## 🔵 Tiny observation (not worth addressing)
`_qdrant_indexes_ensured` gets set to `True` even when `ensure_issuer_id_index` raises (the flag is set after the `except` block, not only on success). A retry within the same process won't re-attempt the index creation. For a single script run that's fine — if it failed once for an environmental reason it'll fail again — and avoiding retry spam in the per-cluster loop is probably the intent.

## Note on the fast path
The post-index `set_payload` calls via `reassign_issuer_on_points` still go through `vectorizer._qdrant()` with the default 5 s timeout, but that's correct now: once the keyword index exists, filtered `set_payload` uses the index and is fast, so 5 s is ample. The long timeout was only needed for the one-time full-scan path and the index build itself.

## Verdict
**Approve, no concerns.** This closes the last latent footgun — the PR is ready to merge.

PR #33, opened 2026-04-16
Fix source link display in research notes

### LINKED ISSUE #30: Research notes - links to source files are not displayed properly
So the links to the source files are not properly displayed. Their names are also a bit long. They include the ticker and they include the date as well as the timestamp. I don't really need the timestamp. Ticker and date in a format is fine

nvasilevv:
## Summary
- Truncate `filing_date` to `YYYY-MM-DD` in citation label construction, so wikilink labels show `TICKER 2026-03-26` instead of `TICKER 2026-03-26T00:00:00+00:00`
- Store date-only (`.date().isoformat()`) in Qdrant metadata for new ingestions, preventing the issue at the source
- The `[:10]` slice in `_label()` also handles existing data that already has full timestamps stored

## Test plan
- [x] All 7 existing `test_cite_translator` tests pass
- [ ] Create a research note via the agent and verify wikilink labels show `TICKER YYYY-MM-DD`
- [ ] Verify existing notes with old timestamp data still render correctly

Closes #30


nvasilevv:
## PR Review — Fix source link display in research notes

### Summary
Clean, minimal fix for a cosmetic bug where wikilink labels showed full ISO timestamps instead of date-only. Two-layer approach: truncation at the display layer for backward compat with existing data, plus date-only storage in Qdrant for new ingestions.

### Findings

**🟡 Improvement (non-blocking)**
- `[:10]` slice in `_label()` is a best-effort truncation rather than a parse-and-format — acceptable since the field is set by your own vectorizer.
- `.date()` call in `vectorizer.py` assumes `filing.filing_date` is always a `datetime`, not a `date` object — worth verifying the ORM column type.
- Duplicate metadata dict construction across `upsert_chunks_from_db` and `upsert_chunks_with_vectors` is pre-existing, not introduced here.

### Positive Feedback
- Pragmatic two-layer fix: source + display.
- Surgical change with no scope creep.
- Test updated to match new behavior.

### Conclusion
**Approve.** No blockers — ship it. ✅

PR #34, opened 2026-04-16
Wire Note.theme_config into WeasyPrint PDF export

nvasilevv:
## Summary

- Resolve per-note `font_family` and `font_size` from `Note.theme_config` in the PDF rendering pipeline so TipTap editor typography preferences carry through to exported PDFs.
- Map the editor's shorthand values (`sans`, `serif`, `mono`) to CSS font stacks via `_THEME_FONT_FAMILIES`.
- `theme_config` overrides branding defaults; when absent, existing behavior is unchanged.
- Template uses a `--font-size` CSS variable instead of a hardcoded `10pt`; header/footer chrome stays fixed at `8pt`.

Addresses item 4 of #17.

## Test plan

- [ ] Generate a PDF report without `theme_config` — verify output matches current behavior (IBM Plex Sans, 10pt)
- [ ] Pass `theme_config={"font_family": "serif", "font_size": "12pt"}` to `render_report()` — verify PDF uses serif font stack at 12pt
- [ ] Pass unknown `font_family` value (e.g. `"Comic Sans"`) — verify it passes through as-is
- [ ] Confirm header/footer text remains at 8pt regardless of `font_size`


nvasilevv:
## Summary

This PR wires `Note.theme_config` typography settings (font family and font size) into the WeasyPrint PDF export pipeline. It adds a mapping layer from TipTap editor shorthand values (`sans`, `serif`, `mono`) to CSS font stacks and uses a CSS variable for dynamic font sizing, while preserving existing branding defaults as fallbacks.

---

## Review Findings

### 🔴 Critical/Security

| # | Finding |
|---|---------|
| 1 | **XSS via `font_size` template injection.** The `font_size` value from `theme_config` is interpolated directly into the CSS `--font-size` variable in `base.html:13` (`{{ font_size | default('10pt') }}`). Since `autoescape=False` is set on the Jinja env (`report_renderer.py:23`), a malicious `theme_config` value like `10pt; } body { display:none } :root { --x:` could inject arbitrary CSS. While this is a PDF renderer (not a browser), WeasyPrint still parses the CSS — a crafted value could break layout or hide content. The `font_family` has the same exposure on line 8. **Recommendation:** Sanitize or validate `font_size` (e.g., regex match `^\d+(\.\d+)?(pt|px|em|rem)$`) and `font_family` before passing to the template. |

### 🟡 Improvement

| # | Finding |
|---|---------|
| 2 | **Caller not wired up.** The `render_report()` function now accepts `theme_config`, but the only call site at `routes.py:2308-2315` does **not** pass `theme_config`. The parameter defaults to `None`, so this is not a runtime error, but the feature is effectively dead code until the route is updated. The PR should either wire the caller or note this as a follow-up. |
| 3 | **`font_family` passthrough could inject CSS too.** `_resolve_theme_font_family` falls back to returning the raw key as-is (`return _THEME_FONT_FAMILIES.get(key, key)`) at `report_renderer.py:43`. Combined with `autoescape=False`, an unknown `font_family` value is injected verbatim into the CSS `--font` variable. This is intentional per the test plan ("unknown value passes through as-is"), but it widens the CSS injection surface. Consider an allowlist or at least strip characters like `}`, `;`, and `{`. |
| 4 | **Falsy `font_size` values.** The expression `(theme_config or {}).get("font_size") or "10pt"` will treat `font_size: ""` (empty string) or `font_size: 0` as falsy and fall back to `"10pt"`. This is probably fine but worth noting — an explicit empty string in theme_config silently gets overridden rather than raising an error. |

### 🔵 Nitpick

| # | Finding |
|---|---------|
| 5 | The `_resolve_theme_font_family` docstring says "Returns None if theme_config is missing or has no font_family" — correct and clear. |
| 6 | The `--font-size` CSS variable name is consistent with the existing `--font` pattern. |

---

## Positive Feedback

- Clean separation: the `_resolve_theme_font_family` helper isolates the mapping logic and keeps `render_report` readable.
- Good fallback chain: theme_config overrides branding, branding overrides hardcoded defaults — no behavior change when `theme_config` is absent.
- The `_THEME_FONT_FAMILIES` mapping with proper CSS font stacks is well-chosen for cross-platform PDF rendering.
- Header/footer chrome correctly stays at fixed `8pt` — only body text is affected.

---

## Conclusion

**Request Changes.**

The CSS injection vector (#1) should be addressed before merging — even in a PDF context, unsanitized user input in CSS is risky. The missing caller wiring (#2) means the feature can't actually be exercised yet, which should be either fixed in this PR or tracked explicitly as a follow-up issue.

nvasilevv:
## Follow-up review

Issues #1, #3, and #4 from the original review are resolved — `font_size` is now regex-validated and `font_family` uses a strict allowlist. Nice fixes.

### Still open

| # | Finding |
|---|---------|
| 2 | **Caller not wired up.** `routes.py:2308-2315` still does not pass `theme_config` to `render_report()`. The parameter defaults to `None`, so no runtime error, but the feature is unreachable from the API. Either wire it here or track it as a follow-up issue. |

nvasilevv:
## Addressing finding #2: caller not wired

The only current caller of `render_report()` is `execute_generate_report` in `routes.py:2308-2315`, which generates reports from **templates** (template + ticker). There's no `Note` object in scope — reports aren't note-based, so passing `theme_config` there would be architecturally wrong.

The right place to wire it is a future note PDF export endpoint (e.g. `GET /notes/{id}/export/pdf`), which would load the note and pass `note.theme_config` to `render_report()`. The plumbing in `render_report()` is already done and sanitized — only the caller is missing.

Tracked as #36.

nvasilevv:
Thanks for the explanation — that makes sense. The renderer-side plumbing is complete and properly sanitized, and wiring `theme_config` into the template-based report caller would be architecturally wrong since there's no `Note` in scope there. Happy with #36 tracking the future note export endpoint.

All original concerns are addressed. LGTM ✅

PR #35, opened 2026-04-16
Fix note editing: sync TipTap editable state on readOnly change

### LINKED ISSUE #29: Note editing
When I try to edit a note I don't geg a keyboard popup on my phone. Neither is the cursor visible to navigate.

nvasilevv:
## Summary

- **Fixes #29** — editing notes was broken on both mobile (no keyboard) and desktop (no cursor)
- `useEditor` only reads the `editable` config at creation time. When toggling an existing note into edit mode, the `TipTapViewer` component key stays the same (`selected?.id`), so the editor instance is never re-created and remains read-only.
- Added a `useEffect` in `TipTapViewer` that calls `editor.setEditable(!readOnly)` when the `readOnly` prop changes, and `editor.commands.focus("end")` to place the cursor and trigger the mobile keyboard.

## Test plan

- [ ] Open the notes page, select an existing note, click **Edit** — editor should become editable with a visible cursor
- [ ] On mobile: keyboard should appear when entering edit mode
- [ ] Type text and verify content updates (onChange fires)
- [ ] Click cancel — editor returns to read-only
- [ ] Create a **new** note — verify it still works as before (regression check)


nvasilevv:
Thanks! Here's what to test after merge:

1. **Desktop — edit existing note:** Open the notes page, select a note, click **Edit**. Verify the editor becomes editable and a blinking cursor appears at the end.
2. **Desktop — type & save:** Type some text, confirm `onChange` fires (content updates). Click cancel — editor should return to read-only.
3. **Mobile — keyboard appears:** On a phone/tablet, open a note and tap **Edit**. The virtual keyboard should pop up automatically.
4. **New note regression:** Create a brand-new note — make sure the flow still works as before (editor should be editable from the start).
5. **Toggle rapidly:** Click Edit → Cancel → Edit a few times in quick succession to confirm no stale state.

PR #38, opened 2026-04-16
Doc Viewer V2: PDF font_size + slash-command menu (#17)

nvasilevv:
## Summary

Addresses the remaining work from #17 (Document Viewer V2 remaining work after PR #14). After investigation, 3 of the 5 items were already implemented — this PR covers the 2 that needed code:

- **Wire `font_size` into PDF rendering**: Added `font_size` field to `ReportBranding` model with init_db migration, plumbed through GET/PATCH branding endpoints and the report generation pipeline into the Jinja2 template as a `--font-size` CSS variable. Defaults to `10pt` for backward compatibility.
- **Slash-command menu**: Implemented a `/` command palette using `@tiptap/suggestion` + `tippy.js` (both already installed but unused). Provides 9 commands: Heading 1/2/3, Bullet List, Ordered List, Code Block, Chart Embed, Citation, Note Link. Vault-node commands (chart/citation/note) reuse the existing `WikilinkAutocomplete` by inserting trigger text (`![[`, `[[`, `[[note_id:`). Only active in editable mode.

### Items confirmed already done
- Alembic migration → handled by `init_db.py` pattern
- WikilinkAutocomplete + TipTap → fully wired in `notes/page.tsx`
- Mark.js Ctrl+F highlighting → works against TipTap DOM via `useTextSearch` hook

## Test plan

- [ ] Generate a PDF report with default branding — verify body text renders at 10pt
- [ ] Set `font_size: "12pt"` on branding via PATCH endpoint — regenerate — verify larger body text
- [ ] Open a note in edit mode, type `/` at start of line — verify popup appears
- [ ] Type `/head` — verify filtering to heading commands
- [ ] Select "Heading 1" — verify heading is applied and `/head` text is removed
- [ ] Type `/chart` and select — verify `![[` is inserted and WikilinkAutocomplete opens
- [ ] Press Escape — verify popup dismisses
- [ ] Verify `/` does NOT trigger popup in read-only mode (chat viewer)


nvasilevv (commented):
## Summary

PR #38 adds two features to the Document Viewer V2: (1) a `font_size` field wired through the full PDF report pipeline (model → migration → API → Jinja2 template), and (2) a TipTap slash-command menu (`/`) for the editable note editor with 9 commands. The changes are well-scoped — 322 additions across 7 files, with the bulk being the new slash-command component.

---

## Review Findings

### 🔴 Critical/Security

- **CSS injection via `font_size` in Jinja2 template** — `backend/app/templates/reports/base.html:13`: The `font_size` value is interpolated directly into a CSS `<style>` block with no sanitization or validation. A user could PATCH `font_size` to something like `10pt; } body { display:none } :root { --x:` or worse, breaking the PDF layout or injecting arbitrary CSS. The `font_family` field has the same pre-existing vulnerability, but `font_size` extends the attack surface. **Recommendation:** Validate `font_size` against a whitelist pattern (e.g., `^\d+(pt|px|em|rem)$`) either in the PATCH endpoint or the model.

- **No input validation on PATCH `/reports/branding`** — `backend/app/api/routes.py:786-801`: The endpoint accepts a raw `dict` body with no schema/Pydantic model, so any string value (including empty, excessively long, or malicious content) is blindly stored via `setattr`. This is pre-existing but now applies to `font_size` too. Consider adding a Pydantic request model with constrained fields.

### 🟡 Improvement

- **`font_size` default repeated in 4 places** — The `"10pt"` default appears in `routes.py` (lines 770, 778, 2145), `report_renderer.py` (line 111), and `base.html` (line 13). If the default ever changes, it's easy to miss one. Consider extracting a constant (e.g., `DEFAULT_FONT_SIZE = "10pt"` in the model or a config module).

- **`startOfLine: true` may surprise users** — `frontend/src/lib/tiptap/slash-command.tsx:238`: The slash command only triggers at the start of a line. This is a deliberate design choice (per PR description), but users commonly expect `/` to work mid-paragraph in tools like Notion. Worth confirming this is intentional UX.

- **Tippy.js popup cleanup on editor destroy** — The `onExit` handler (`slash-command.tsx:277-281`) cleans up correctly, but if the editor is destroyed while the popup is open (e.g., navigating away), the tippy instance could leak. The TipTap suggestion plugin typically handles this, but worth a quick manual test.

### 🔵 Nitpick

- **`type` cast on tippy return** — `slash-command.tsx:262`: `as unknown as TippyInstance` is a double cast. Tippy's `tippy()` on `document.body` returns an `Instance` directly when given a single element — the cast suggests a type mismatch that could be resolved with a proper import or by using `tippy(document.body, ...)[0]` if it returns an array.

---

## Positive Feedback

- **Clean plumbing of `font_size`** — The CSS variable approach (`--font-size`) is elegant and consistent with the existing `--font` / `--primary` pattern. Backward compatibility via `"10pt"` default is well handled.
- **Slash-command architecture is solid** — Using `@tiptap/suggestion` + tippy.js is the canonical approach. The `forwardRef` + `useImperativeHandle` pattern for keyboard handling is correct. The trick of reusing `WikilinkAutocomplete` by inserting trigger text (`![[`, `[[`, `[[note_id:`) for vault-node commands is clever and avoids duplicating autocomplete logic.
- **Conditional extension loading** — `...(!readOnly ? [SlashCommands] : [])` in `tiptap-viewer.tsx:105` is a clean way to disable the menu in read-only mode.
- **Init_db migration with idempotent `IF NOT EXISTS`** — Follows the established pattern in the codebase.

---

## Conclusion

**Request Changes** — The CSS injection vector through unsanitized `font_size` in the Jinja2 template is the blocking issue. Add a validation regex (e.g., `^\d{1,3}(pt|px|em|rem)$`) in the PATCH endpoint before merging. The rest of the PR is clean and well-structured.

nvasilevv (commented):
The new commit (`7f04782`) addresses the critical and main improvement findings:

**Resolved:**
- **CSS injection** — `_FONT_SIZE_RE` regex (`^\d{1,3}(pt|px|em|rem)$`) now validates `font_size` in the PATCH endpoint before storage, returning a 400 on invalid input. Injection vector is closed.
- **Repeated default** — `DEFAULT_FONT_SIZE = "10pt"` extracted to `report_branding.py` and used across `routes.py`. Clean.

**Minor note:** `report_renderer.py:111` still has a literal `"10pt"` fallback (`or "10pt"` instead of `or DEFAULT_FONT_SIZE`) — not blocking since it's a defense-in-depth fallback, but worth aligning for consistency.

The remaining items from the initial review (raw `dict` body on PATCH, `startOfLine` UX, tippy cleanup, type cast) were non-blocking and can be addressed separately.

Looks good to merge. 👍

Issue #19, opened 2026-04-14
Notes editor drops markdown tables: TipTap Table extension not registered


nvasilevv:
## Summary

Markdown tables produced by the agent's `create_note` tool are stored correctly in the database but render as a single run of concatenated cell text in the Research Notes editor — the table grid is missing entirely.

## Reproduction

Live note `4c804fe8-b36f-474e-b92e-3366d74fa94e` ("Shelly Group — Growth Investor Research Note (2020–2025)") contains a GFM pipe table under `## Financial snapshot`:

```
| Metric | 2020 | 2021 | 2022 | 2023 | 2024 |
|---|---:|---:|---:|---:|---:|
| Revenue (BGN m) | 46.3 | 59.5 | 93.2 | 146.5 | 208.7 |
...
```

What the user sees in the Research Notes tab:

> Metric20202021202220232024Revenue (BGN m)46.359.593.2146.5208.7EBIT / Operating profit (BGN m)12.418.720.437.450.5…

No borders, no rows, no columns — all cell text collapsed into one line.

## Root cause

`frontend/src/components/tiptap/tiptap-viewer.tsx:96-105` registers only `StarterKit`, `Markdown` (`tiptap-markdown`), and `VaultWikilinkNodes`. StarterKit does **not** include table support — TipTap v3 ships Table/TableRow/TableHeader/TableCell as separate `@tiptap/extension-table*` packages, and none of them are present in `frontend/package.json`.

`tiptap-markdown` (via markdown-it) still tokenizes the pipe table, but with no matching ProseMirror node to land on, the cell text falls into inline content with no structure — hence the concatenation.

The backend is innocent: the DB round-trip is verified byte-for-byte identical to what the agent emitted (minus the expected `[cite-N]` → `[[document:…]]` rewrite). The table block survives persistence and is present verbatim in `note_revision.content`.

## Fix

1. Add deps to `frontend/package.json`:
   - `@tiptap/extension-table`
   - `@tiptap/extension-table-row`
   - `@tiptap/extension-table-header`
   - `@tiptap/extension-table-cell`
2. Register them in the `extensions` array in `frontend/src/components/tiptap/tiptap-viewer.tsx`:
   ```ts
   Table.configure({ resizable: false }),
   TableRow,
   TableHeader,
   TableCell,
   ```
[...]

PR #39, opened 2026-04-16
Preserve column alignment in markdown table round-trip

### LINKED ISSUE #27: Research notes - issue with mardown tables display
Markdown tables are not parsed properly.

nvasilevv:
## Summary

Closes #27.

- `tiptap-markdown` v0.9.0 hardcodes `---` for every delimiter column when serializing tables back to markdown, silently dropping GFM alignment markers (`:---`, `---:`, `:---:`) after a single edit-save cycle
- Added a custom `TableWithMarkdown` extension (`frontend/src/lib/tiptap/table-markdown.ts`) that overrides the built-in serializer to read each header cell's `align` attribute and emit the correct delimiter
- Complex tables (merged cells, multi-paragraph content) still fall back to HTML serialization

Depends on #19 (table extension registration), which is already merged.

## Test plan

- [ ] Create a note with an aligned GFM table (`:---|:---:|---:` delimiters) and verify alignment renders correctly
- [ ] Edit the note, save, and verify the markdown preserves alignment markers on round-trip
- [ ] Test tables with empty cells, bold/italic content, and wikilinks in cells
- [ ] Run `npx tsc --noEmit` in `frontend/` — passes clean


nvasilevv (commented):
## PR Review: Preserve column alignment in markdown table round-trip (#39)

### Summary
This PR adds TipTap table extension support and a custom markdown serializer (`TableWithMarkdown`) that preserves GFM column alignment markers (`:---`, `---:`, `:---:`) during edit-save cycles. The upstream `tiptap-markdown` serializer hardcodes `---` for all delimiters, silently dropping alignment — this override reads each header cell's `align` attribute and emits the correct delimiter. Complex tables (merged cells, multi-paragraph content) gracefully fall back to HTML serialization.

---

### Review Findings

#### 🔴 Critical/Security
- **None found.** No injection vectors, no secrets, no auth concerns. The four new `@tiptap/extension-table-*` dependencies are official TipTap packages from `@tiptap` org — no supply-chain risk.

#### 🟡 Improvement

- **HTML fallback XSS surface** (`table-markdown.ts:143-155`): `serializeTableHtml` interpolates `cell.textContent` directly into HTML tags. While `textContent` strips HTML by definition (ProseMirror text nodes are plain text), if this ever changes to `innerHTML` or processes rich inline content, it becomes an XSS vector. Consider adding a brief comment noting this assumption, or escaping `<` / `>` / `&` in the output as a defensive measure.

- **Single editor instance** — The table extensions are only registered in `tiptap-viewer.tsx`. Currently that's the only editor in the codebase, so this is fine. But if a second editor is ever introduced, the table support (and alignment preservation) would need to be duplicated. Minor — just worth noting for future awareness.

- **`state.inTable` flag** (`table-markdown.ts:98, 112`): The serializer sets `state.inTable = true` but doesn't guard against re-entrant calls (nested tables). GFM doesn't support nested tables so this is unlikely, but if a malformed document somehow nests tables, `inTable` would be left `true` after the inner table's `closeBlock`. Low risk given the `isMarkdownSerializable` guard, but a `try/finally` around the block would make it bulletproof.

#### 🔵 Nitpick

- **Placeholder URL in JSDoc** (`table-markdown.ts:13`): `@see https://github.com/your-org/chewier/issues/27` — should be `https://github.com/nvasilevv/chewier/issues/27`.

- **Unused parameter names** (`table-markdown.ts:100, 103`): `_p` and `_p2` (position offsets) — could use just `_` for both since they're intentionally unused, though this is purely stylistic.

---

### Positive Feedback

- **Clean separation of concerns**: The `TableWithMarkdown` extension cleanly extends `@tiptap/extension-table` via `addStorage()` without monkey-patching or importing tiptap-markdown internals. This is the correct integration point and will survive upstream upgrades.

- **Robust fallback logic**: `isMarkdownSerializable()` correctly identifies tables that can't be losslessly represented as GFM (merged cells, multi-paragraph content) and falls back to HTML. This avoids data loss for edge cases rather than silently mangling them.

- **Well-structured helpers**: `childNodes`, `hasSpan`, `alignDelimiter`, and `isMarkdownSerializable` are small, focused, and easy to reason about. The overall file reads like a well-documented module.

- **Correct dependency choices**: Using the official `@tiptap/extension-table-*` packages at the same version (`3.22.3`) as the existing TipTap packages avoids version skew.

---

### Conclusion

**Approve** — The code is clean, well-structured, and solves a real data-loss bug. The improvements noted above are minor hardening suggestions, none of which block merging.

nvasilevv:
## Follow-up review after `a1f5777`

| Review Item | Status |
|---|---|
| 🔵 **Placeholder JSDoc URL** (`your-org` → `nvasilevv`) | ✅ Fixed (line 13) |
| 🟡 **`state.inTable` re-entrancy** — wrap in `try/finally` | ✅ Fixed (lines 100-115) |
| 🟡 **HTML fallback XSS surface** — add escaping or comment | Not addressed — `cell.textContent` is still interpolated directly. Low risk since ProseMirror `textContent` is plain text by definition, but the defensive comment/escaping wasn't added. |
| 🔵 **Unused `_p`/`_p2` parameter names** | Not addressed — purely stylistic, fine to leave. |

Two of the four items were addressed. The remaining two are a low-risk suggestion and a stylistic nitpick — neither blocks merging. LGTM 👍

PR #41, opened 2026-04-16
Fix font size changes not applied in notes editor

### LINKED ISSUE #37: Font size changes not applied in notes editor
## Description

Changing the font size in the notes editor has no visible effect. The control appears functional but the text size doesn't update.

## Steps to reproduce

1. Open the notes page and select a note
2. Click **Edit** to enter edit mode
3. Select text and attempt to change the font size
4. Observe that the font size does not change

## Expected behavior

Selected text should update to the chosen font size.

## Context

Note editing itself was fixed in #35 (TipTap editable sync). This issue may be related to a missing TipTap extension or a CSS override.

nvasilevv:
## Summary

Closes #37

- The font size dropdown in the notes editor had no visible effect because the inline `fontSize` style was applied on a **parent** wrapper div, while the child `.prose-ceewire` CSS class set `font-size: 0.875rem` directly — a directly-applied rule always wins over an inherited value.
- Added an optional `style` prop to `TipTapViewer` and moved the `fontFamily`/`fontSize` inline style onto the `.prose-ceewire` root element, where it correctly overrides the CSS class.
- Other consumers of `.prose-ceewire` (document viewer, ask panel, data tables) are unaffected — they don't pass a `style` prop and keep the default `0.875rem`.

## Test plan

- [ ] Open the notes page, select a note, click **Edit**
- [ ] Change the font size dropdown — text should visually update immediately
- [ ] Change the font family dropdown — text should update
- [ ] Save the note, re-open — theme settings should persist
- [ ] Verify document viewer and ask panel still render at their default font size


nvasilevv:
## PR #41 Review: Fix font size changes not applied in notes editor

### Summary

This PR fixes issue #37 where the font size dropdown in the notes editor had no visible effect. The root cause was that the inline `fontSize` style was applied on a parent wrapper `<div>`, but the child `.prose-ceewire` CSS class set `font-size: 0.875rem` directly on the element — a directly-applied CSS rule always beats an inherited value. The fix moves the inline style onto the `TipTapViewer` root element itself (which carries the `.prose-ceewire` class), so the inline style correctly overrides the CSS class rule.

### Review Findings

**No critical or security issues found.**

| Severity | Finding |
|----------|---------|
| 🟡 **Improvement** | The heading sizes in `.prose-ceewire` (`h1: 1.375rem`, `h2: 1.2rem`, etc.) at `globals.css:121-124` are absolute `rem` values. When the user changes `fontSize` via the dropdown, body text will scale but headings won't — they'll stay fixed. Consider converting heading sizes to `em` units (relative to the parent's font-size) so they scale proportionally with the user's chosen font size. This is a pre-existing issue, not introduced by this PR. |
| 🟡 **Improvement** | Same applies to `table` and `code` font sizes (`.prose-ceewire table { font-size: 0.8125rem }`, `.prose-ceewire code { font-size: 0.8125rem }`) — these won't respond to the user's font-size preference either. Again, pre-existing. |
| 🔵 **Nitpick** | The `style` prop is placed between `readOnly` and `onChange` in the JSX (`notes/page.tsx:458-461`). Grouping it next to `className` would be marginally more conventional since both are presentation-related, but this is purely cosmetic. |

### Positive Feedback

- **Correct root-cause analysis.** The PR description clearly explains *why* CSS inheritance didn't work (directly-applied class rule vs. inherited inline style). This is well-documented.
- **Minimal, surgical change.** Only 2 files touched, 8 additions / 7 deletions. The fix adds a generic `style` prop to `TipTapViewer` rather than a font-specific prop, which is the right abstraction level for a reusable component.
- **No impact on other consumers.** The `style` prop is optional and defaults to `undefined`, so the document viewer, ask panel, and data tables are completely unaffected.
- **Good use of the existing `cn()` pattern** — the component already merges `className` via `cn()`, and the new `style` prop follows the same passthrough convention.

### Conclusion

**Approve.** The fix is correct, minimal, and well-scoped. The CSS specificity issue is properly resolved by applying the inline style at the same element that carries the `.prose-ceewire` class. The improvement suggestions around heading/table/code `rem` vs `em` units are pre-existing and can be addressed in a follow-up if desired.

nvasilevv:
**Review: Approve** ✅

Root cause analysis is correct: inline `fontSize` on the parent wrapper was inherited, but `.prose-ceewire` sets `font-size: 0.875rem` directly on the child — a directly-applied rule always wins over an inherited value. Moving the inline `style` onto the `.prose-ceewire` root element inside `TipTapViewer` is the right fix.

No side effects on other consumers (document viewer, ask panel, data tables) since the `style` prop is optional and they don't pass it.

Clean, minimal change — 8 additions, 7 deletions across 2 files.

PR #42, opened 2026-04-16
Fix slash commands not appearing when editing existing notes

### LINKED ISSUE #40: Note editor - slash commands work when creating from scratch but not when editing

nvasilevv:
## Summary

Fixes #40.

- **Problem:** Slash commands (triggered by typing `/`) worked when creating a new note but not when editing an existing one.
- **Root cause:** `SlashCommandExtension` was conditionally included in the TipTap extensions array based on the `readOnly` prop at mount time. When viewing an existing note (read-only) and then clicking "Edit", the editor instance kept its original extensions — the slash command plugin was never registered.
- **Fix:** Always include `SlashCommandExtension`. TipTap's Suggestion plugin already checks `editor.isEditable` internally, so it stays inert in read-only mode while being immediately available when the user toggles to edit.

## Test plan

- [ ] Create a new note — type `/` and confirm slash commands appear
- [ ] Save the note, click "Edit" — type `/` and confirm slash commands appear
- [ ] View a note in read-only mode — confirm slash menu does not appear


nvasilevv (commented):
## Summary

This single-file PR fixes slash commands (`/`) not appearing when editing an existing note. The root cause was that `SlashCommandExtension` was conditionally included in TipTap's extensions array based on `readOnly` at mount time — when a user viewed a note (read-only) and then clicked "Edit," the editor kept its original extensions and never registered the slash menu. The fix includes the extension unconditionally, relying on TipTap's built-in `editor.isEditable` check.

## Review Findings

### 🔴 Critical/Security
- **None found.** No injection vectors, no secrets, no auth changes, no new dependencies.

### 🟡 Improvement
- **Verify the `editor.isEditable` claim.** The comment states that TipTap's Suggestion plugin checks `editor.isEditable` internally. This is correct for TipTap's `@tiptap/suggestion` (it checks `editor.isEditable` in the `handleKeyDown` and decoration logic), but worth a quick smoke test: open a note in read-only mode and press `/` — confirm no popup appears. If your `SlashCommandExtension` wraps a custom trigger or overrides `allow()`, a slash menu appearing in read-only mode would be a usability bug.
- **Old comment was technically inaccurate anyway.** The removed comment said _"both current callers mount with a fixed value"_ — but the notes page clearly toggles `readOnly` at runtime (view -> edit). Removing it was the right call; the new comment is accurate.

### 🔵 Nitpick
- **None.** The change is minimal and well-scoped.

## Positive Feedback

- **Excellent root cause analysis** in the PR description — it clearly explains the problem, why it happens, and why the fix is safe. This is a model PR description.
- **Minimal, surgical fix** — 4 additions, 5 deletions, single file. No unrelated changes bundled in.
- **Leverages framework behavior** instead of adding custom workaround code (e.g., dynamically registering/unregistering extensions). This is the cleanest possible fix.

## Conclusion

**Approve** — contingent on confirming the test plan items (slash menu hidden in read-only mode, visible in both new-note and edit-existing-note flows). The fix is correct, minimal, and well-reasoned.

PR #43, opened 2026-04-16
Add note PDF export endpoint with theme_config support

### LINKED ISSUE #36: Wire theme_config to render_report caller when note PDF export is added
## Context

PR #34 added `theme_config` support to `render_report()` in `report_renderer.py`, with sanitized `font_family` (strict allowlist) and `font_size` (regex-validated) resolution.

However, the only current caller — `execute_generate_report` in `routes.py:2308-2315` — generates reports from **templates** (template + ticker), not from notes. There is no `Note` object in scope to pull `theme_config` from, so passing it there would be architecturally wrong.

## What needs to happen

When a **note PDF export** endpoint is added (e.g. `GET /notes/{id}/export/pdf`), it should:

1. Load the note and its `theme_config`
2. Pass `theme_config=note.theme_config` to `render_report()`

The plumbing in `render_report()` is already done and validated — this is purely about adding the caller.

## Related

- PR #34 (theme_config → PDF pipeline)
- Issue #17 item 4 (WeasyPrint PDF round-trip)

nvasilevv:
## Summary

- Adds `GET /notes/{note_id}/export/pdf` endpoint that renders a note as a downloadable PDF
- Introduces `render_note_pdf()` in `report_renderer.py`, reusing existing markdown, font resolution, and chart rendering helpers
- Creates `note.html` Jinja2 template (extends `base.html`) with a note-appropriate cover page (title + type, no ticker/company)
- Passes `Note.theme_config` through to override branding font defaults (font_family, font_size)
- Supports optional `?revision=N` query param to export a specific revision
- Pre-fetches chart definitions and linked notes for wikilink rendering in the PDF
- Streams the PDF directly (no R2 storage) since notes are living documents

Closes #36

## Test plan

- [ ] Create a note with markdown content and verify `GET /notes/{id}/export/pdf` returns a valid PDF
- [ ] Set `theme_config` (e.g. `{"font_family": "serif", "font_size": "12pt"}`) and verify fonts apply in the exported PDF
- [ ] Export without `theme_config` — should use user's ReportBranding defaults
- [ ] Test `?revision=1` param exports the correct revision
- [ ] Verify 404 for non-existent note, unauthorized note, and missing revision
- [ ] Verify notes with `![[chart:UUID]]` and `[[note_id:UUID|Title]]` wikilinks render correctly


nvasilevv:
## PR Review — Add note PDF export endpoint with theme_config support

### Summary

This PR adds a `GET /notes/{note_id}/export/pdf` endpoint that renders a note as a downloadable PDF, reusing the existing report rendering infrastructure. It also wires `font_size` into the `ReportBranding` model (new DB column, GET/PATCH endpoints) and uses a new `note.html` Jinja2 template with a note-appropriate cover page. The changes are well-scoped and follow the existing patterns in `report_renderer.py`.

---

### Review Findings

#### 🔴 Critical/Security

1. **`font_size` validation bypass on `None`** — `backend/app/api/routes.py:797-798`
   ```python
   if "font_size" in body and not _FONT_SIZE_RE.match(body["font_size"] or ""):
   ```
   If `body["font_size"]` is `None`, the `or ""` makes the match fail on an empty string — so `None` raises a 400. But `font_size` is `Optional[str]` on the model, and the user should be able to clear it by sending `null`. Either allow `None` explicitly (`if body.get("font_size") is not None and not _FONT_SIZE_RE.match(...)`) or document that null clears the override.

2. **Missing top-level imports in endpoint** — The `export_note_pdf` endpoint has heavy inline imports (`from app.models.report_branding import ...`, `from app.services.report_renderer import ...`) executed on every request. While Python caches module imports, this is a code smell. Consider moving them to the top of the file, consistent with how `ChartDefinition` and `Note` are already imported at module level.

3. **No `Content-Disposition` filename encoding** — `backend/app/api/notes.py:498`
   ```python
   headers={"Content-Disposition": f'attachment; filename="{safe_title}.pdf"'},
   ```
   The `safe_title` sanitization replaces non-alphanumeric chars with `_`, but doesn't handle quotes in the title and doesn't use RFC 5987 `filename*=UTF-8''...` encoding for non-ASCII characters. Since note titles can contain Unicode, the download filename could be garbled or malformed. Consider using `urllib.parse.quote` for the `filename*` parameter.

#### 🟡 Improvement

4. **Duplicated branding dict construction** — The branding dict is built identically in three places: `export_note_pdf` (notes.py:444-451), `get_report_branding` (routes.py:765-776), and `execute_generate_report` (routes.py:2141-2149). Extract a helper like `build_branding_dict(row)` to keep this DRY. If a new field is added to `ReportBranding`, only one of these three locations is likely to be updated.

5. **`render_note_pdf` font_size fallback differs from branding default** — `report_renderer.py:211`
   ```python
   font_size = raw_size if _VALID_FONT_SIZE.match(raw_size) else "10pt"
   ```
   This hardcodes `"10pt"` instead of using `DEFAULT_FONT_SIZE` (which is also `"10pt"` today but may drift). The `render_report` function has the same hardcoded fallback at line 136 — both should reference the constant.

6. **Branding `font_size` not passed through from note's `theme_config`** — In `export_note_pdf` (notes.py:480-487), the branding dict's `font_size` is assembled but `render_note_pdf` resolves `font_size` from `theme_config` only, falling back to `"10pt"`. The branding-level `font_size` is never used as a fallback in `render_note_pdf`. Is that intentional? If users set a custom `font_size` on their branding but have no `theme_config`, they'll get `10pt` instead of their branding preference.

7. **The endpoint buffers the full PDF, not streams** — The PR description says "streams the PDF directly" but `Response(content=pdf_bytes)` returns the entire buffer. This is fine for typical note sizes, but the description is misleading.

#### 🔵 Nitpick

8. **Route ordering** — The `export/pdf` endpoint is placed between `list_revisions` and `delete_note`. Grouping all `{note_id}` sub-routes together (get, update, revisions, export, delete) would improve readability.

9. **`note.html` template missing escaping for `note_type`** — `note.html:12`
   ```html
   <div class="meta">{{ note_type | default('') }} &middot; {{ report_date | default('') }}</div>
   ```
   `note_type` comes from user input. Since this is a PDF rendered by WeasyPrint (not served as HTML to a browser), XSS risk is low, but injecting arbitrary HTML into the PDF cover page is still possible. Consider escaping or validating `note_type` against an allowlist.

---

### Positive Feedback

- Excellent reuse of `_markdown_to_html`, `_resolve_theme_font_family`, and chart/wikilink rendering — no duplication of the markdown pipeline.
- The `_FONT_SIZE_RE` regex validation on the PATCH endpoint is a solid defense against CSS injection.
- Pre-fetching chart definitions and linked notes in the endpoint (rather than sync DB access in the renderer) follows the established pattern cleanly.
- The `DEFAULT_FONT_SIZE` constant extraction addresses a real maintenance concern.
- The DB migration uses `IF NOT EXISTS` — safe for idempotent reruns.

---

### Conclusion

**Request Changes** — The `font_size` validation bypass on `None` (item 1) should be resolved before merge, and the duplicated branding dict (item 4) is worth addressing now before it becomes a four-way copy-paste. The rest are improvements that can be tackled in a follow-up.

nvasilevv:
## PR Review: Add note PDF export endpoint with theme_config support

### Summary

This PR adds a `GET /notes/{note_id}/export/pdf` endpoint that renders a note as a downloadable PDF, reusing existing markdown/chart/font infrastructure from the report renderer. It also wires `font_size` into the `ReportBranding` model with a DB migration, input validation, and the existing report generation pipeline. The changes are well-scoped and build cleanly on existing patterns.

---

### Review Findings

#### 🔴 Critical/Security

| # | Finding |
|---|---------|
| 1 | **`content_html \| safe` in `note.html:16` — potential XSS via stored markdown.** The `_markdown_to_html()` function converts user-authored markdown to HTML, then `{{ content_html \| safe }}` renders it unescaped. If the markdown renderer doesn't sanitize HTML tags (e.g. `<script>`, ``), a user could inject JS. This is a PDF context (WeasyPrint won't execute JS), so actual exploitation is very low risk, but if this HTML is ever reused in a browser context it becomes critical. Worth confirming `_markdown_to_html` strips dangerous tags, or adding a note that this is PDF-only. |
| 2 | **`logo_base64` rendered unsanitized in `note.html:8`.** `{{ logo_base64 }}` is injected into a `src="data:image/png;base64,..."` attribute without Jinja2 autoescaping (the env has `autoescape=False`). If an attacker could control the logo storage key to return crafted content, they could break out of the attribute. Low risk since logo upload is user-controlled and WeasyPrint doesn't execute JS, but the `autoescape=False` env is worth flagging. |

#### 🟡 Improvement

| # | Finding |
|---|---------|
| 1 | **Duplicated branding defaults across endpoints.** The branding dict construction (primary_color, font_family, font_size, etc.) with fallback defaults is now copy-pasted in 3 places: `notes.py:export_note_pdf`, `routes.py:get_report_branding`, and `routes.py:execute_generate_report`. Consider extracting a `load_user_branding(session, email) -> dict` helper to keep defaults consistent. |
| 2 | **`render_note_pdf` fallback font_size uses hardcoded `"10pt"` instead of `DEFAULT_FONT_SIZE`.** At `report_renderer.py:213`, the fallback is `"10pt"` as a string literal. The constant `DEFAULT_FONT_SIZE` exists specifically for this — use it for consistency. Same pattern exists in the existing `render_report` at line 136. |
| 3 | **`font_size` validation gap in `routes.py:797`.** If `body["font_size"]` is not a string (e.g. an integer `14`), `_FONT_SIZE_RE.match(body["font_size"] or "")` will throw `TypeError`. The `body` is an untyped `dict` — consider `str(body.get("font_size", ""))` or using a Pydantic model for the request body. |
| 4 | **No `read_only` or `visibility` check beyond `get_note`.** The `get_note` call checks visibility, which is good. But there's no explicit org-level permission check for PDF export specifically — confirm that `notes_service.get_note` is sufficient access control for exporting content. |

#### 🔵 Nitpick

| # | Finding |
|---|---------|
| 1 | **`note.html` cover page uses `\| default('Untitled Note')` but caller already passes `note.title or "Untitled"`.** The template default will never trigger — the two defaults are also inconsistent ("Untitled Note" vs "Untitled"). Pick one. |
| 2 | **`_FONT_SIZE_RE` in routes.py vs `_VALID_FONT_SIZE` in report_renderer.py.** Routes: `^\d{1,3}(pt\|px\|em\|rem)$` (integer only, max 3 digits). Renderer: `^\d+(\.\d+)?(pt\|px\|em\|rem)$` (allows decimals, unlimited digits). The renderer is more permissive. These should ideally be the same regex. |

---

### Positive Feedback

- **Clean reuse of existing infrastructure.** `render_note_pdf` follows the same pattern as `render_report` and reuses `_markdown_to_html`, `_resolve_theme_font_family`, chart/wikilink rendering — no reinvention.
- **Good migration approach.** The `IF NOT EXISTS` guard in `init_db.py` is safe for idempotent deploys.
- **Font size validation with regex** on the PATCH endpoint is a good defense against CSS injection.
- **Streaming the PDF directly** instead of uploading to R2 is the right call for living documents.
- **Pre-fetching chart definitions and note links** before rendering avoids N+1 queries during markdown conversion.

---

### Conclusion

**Comment** — The PR is solid and well-structured. No blocking security issues given the PDF-only rendering context. Main actionable items:

1. Extract the branding defaults into a shared helper (🟡 #1) to prevent drift
2. Use `DEFAULT_FONT_SIZE` instead of `"10pt"` literal in the renderer (🟡 #2)
3. Guard `font_size` validation against non-string input (🟡 #3)

With those addressed, this is an **Approve**.

nvasilevv:
## Follow-up on review

Thanks for addressing the `DEFAULT_FONT_SIZE` constant extraction — that looks good now.

A few items from the original review are still open:

**🟡 `font_size` validation still crashes on non-string input** (`routes.py:797`)
`_FONT_SIZE_RE.match(body["font_size"] or "")` will raise `TypeError` if `font_size` is sent as an integer (e.g. `{"font_size": 14}`). Quick fix:
```python
if "font_size" in body and not _FONT_SIZE_RE.match(str(body["font_size"] or "")):
```

**🟡 Branding dict still duplicated in 3 places**
`notes.py:export_note_pdf`, `routes.py:get_report_branding`, and `routes.py:execute_generate_report` all build the same fallback dict. A shared `load_user_branding(session, email) -> dict` helper would prevent future drift.

**🔵 Minor — inconsistent template defaults**
`note.html` has `{{ note_title | default('Untitled Note') }}` but the caller always passes `note.title or "Untitled"`. The template default never triggers, and the two strings don't match.

nvasilevv:
All three items are addressed in 7522d97:

1. **`font_size` non-string crash** — Guarded with `isinstance(body["font_size"], str)` before the regex match, so integers/nulls get a clean 400 instead of a `TypeError`.

2. **Branding dict duplication** — Extracted `build_branding_dict(row)` in `report_branding.py` and replaced all 3 call sites (`export_note_pdf`, `get_report_branding`, `execute_generate_report`).

3. **Inconsistent template default** — `note.html` now uses `'Untitled'` to match the caller.

Also snuck in two related fixes:
- Unified `_FONT_SIZE_RE` to allow decimals (e.g. `10.5pt`), matching `_VALID_FONT_SIZE` in the renderer
- Renderer now falls back to `branding.font_size` before `DEFAULT_FONT_SIZE`, so users' branding preference applies even without `theme_config`

PR #45, opened 2026-04-16
Fix slash-command search not working in edit mode (#44)

### LINKED ISSUE #44: Slash commands appear after #42 in edit mode as well, but search doesn't work
So, slash commands appear, but when is used to pull the search component (where one can search for a note if /note is used) the search doesn't appear. It does appear when a new note is created.

nvasilevv:
## Summary

Fixes #44 — slash commands appear after #42, but the WikilinkAutocomplete search popup doesn't show when editing an existing note (works fine for new notes).

**Root cause — two bugs:**

- `onEditorReady={editing ? setEditor : undefined}` meant the editor instance was never captured when a note was first opened in read-only mode. Toggling to edit didn't remount `TipTapViewer` (same `key`), so the callback never fired and autocomplete transaction detection never subscribed.
- `SlashCommandExtension` was conditionally excluded at mount time when `readOnly=true`. `useEditor` doesn't rebuild extensions on prop changes, so the extension was permanently missing for existing notes.

**Fix:**
- Always pass `onEditorReady={setEditor}` so the editor instance is always captured
- Always include `SlashCommandExtension` — TipTap's Suggestion plugin already checks `editor.isEditable` internally, so it stays inert in read-only mode

## Test plan

- [ ] Open an existing note → click Edit → type `/` → slash menu appears
- [ ] Select `/note` → WikilinkAutocomplete search popup appears
- [ ] Create a new note → verify slash commands + search still work (regression)
- [ ] View a note in read-only mode → verify no slash menu appears (regression)


nvasilevv (commented):
## PR Review

**Summary:** Clean, minimal fix that correctly addresses both root causes — the conditional `onEditorReady` callback and the conditionally excluded `SlashCommandExtension`. Both bugs stem from `useEditor` not rebuilding extensions on prop changes, and the fix works with that constraint rather than against it.

### Review Findings

| Severity | Finding |
|---|---|
| 🔴 **Critical/Security** | None |
| 🟡 **Improvement** | None — the change is already minimal and correct |
| 🔵 **Nitpick** | None |

### Positive Feedback
- Excellent root cause analysis in the PR description — clearly identifies both bugs and explains *why* the fix works
- The updated comment in `tiptap-viewer.tsx` explains the `useEditor` limitation well, which will save future developers from re-introducing the conditional gate
- Surgical fix: 6 lines changed, no unnecessary refactoring
- Relying on TipTap's built-in `editor.isEditable` check in the Suggestion plugin is the right call — no need for a custom guard

### Functional Correctness
- ✅ Edit existing note → `/` → slash menu appears (editor instance always captured)
- ✅ New note → slash commands still work (no regression)
- ✅ Read-only mode → no slash menu (Suggestion plugin checks `isEditable` internally)
- ✅ Rapid edit toggle — `useEffect` syncs `setEditable`, editor ref always captured

**Verdict: Approve** ✅

PR #49, opened 2026-04-16
Wire up inline chart-link click handling in notes editor

### LINKED ISSUE #16: Inline chart-link clicks in notes editor are a no-op
## Context
Introduced alongside the TipTap editor migration in #14. `ChartLinkNode` (inline `[[chart:UUID|title]]`) now round-trips correctly through the editor, but clicks on those nodes don't do anything.

## Details
`TipTapViewer`'s `handleClick` iterates `NODE_CLICK_SELECTORS` (`frontend/src/components/tiptap/tiptap-viewer.tsx`), which only covers `[data-filing-id]`, `[data-document-id]`, and `[data-note-id]`. The chart-link node renders with `data-chart-link-id` but no selector matches it, and `ClickedVaultNode` has no `"chart"` variant either.

Result: clicking an inline chart reference in a note is a no-op. The embed case (`![[chart:…]]`) is unaffected — that's a block-level NodeView rendering the live `ChartRenderer`.

## Suggested fix
- Extend the `ClickedVaultNode` union with a `chart` variant (`chartId`, `displayText`).
- Add `{ selector: "[data-chart-link-id]", build: … }` to `NODE_CLICK_SELECTORS`.
- Decide the target behavior in `notes/page.tsx` `handleNodeClick` — probably opens a chart preview panel or navigates to the chart's source note.

## Priority
Low — non-regression relative to pre-TipTap behavior (the old renderer styled chart references as spans but also didn't route clicks). Nice to close out for consistency with filing/document/note links.

nvasilevv:
## Summary

Closes #16

- Added `chart` variant to the `ClickedVaultNode` union type in `tiptap-viewer.tsx`
- Added `[data-chart-link-id]` selector to `NODE_CLICK_SELECTORS` so clicks on inline `[[chart:UUID]]` references are detected
- Added `chart` case to `handleNodeClick` in the notes page, opening the charts page in a new tab

The chart-link node already rendered with `data-chart-link-id` but nothing matched it on click. Filing, document, and note links all worked — chart was the only missing variant.

## Test plan

- [ ] Open a note containing an inline `[[chart:UUID|title]]` reference
- [ ] Click the chart link — should open the charts page in a new tab
- [ ] Verify filing, document, and note links still work as before
- [ ] Verify chart embeds (`![[chart:UUID]]`) are unaffected


nvasilevv (commented):
## Summary

This PR wires up click handling for inline `[[chart:UUID]]` wikilink references in the notes editor. The `ClickedVaultNode` union type gains a `chart` variant, a new `[data-chart-link-id]` selector is registered, and the notes page handles chart clicks by opening the charts page in a new tab. Small, focused change that fills a gap — chart links rendered but were inert.

---

## Review Findings

### 🔴 Critical/Security

None.

### 🟡 Improvement

| # | File | Finding |
|---|------|---------|
| 1 | `notes/page.tsx:251` | **Chart link doesn't deep-link to the specific chart.** `window.open('../charts', '_blank')` opens the charts listing page, not the chart identified by `clicked.chartId`. The `chartId` is extracted but unused. Filing/document links resolve to a specific document viewer — chart should do the same. Consider ``window.open(`../charts?id=${clicked.chartId}`, '_blank')`` or whatever selection mechanism the charts page supports. |
| 2 | `notes/page.tsx:250` | **Guard `!clicked.chartId` returns silently after `chartId` defaulted to `""`.** In `tiptap-viewer.tsx:96`, `chartId` is `el.dataset.chartLinkId ?? ""`, so a missing `data-chart-link-id` attribute becomes `""` (falsy). The guard catches this, but the user sees nothing — no toast, no console warning. Other link kinds (note, filing) also fail silently, so this is consistent, but worth considering a user-facing hint at some point. |
| 3 | `notes/page.tsx:251` | **Relative path `../charts` is fragile.** It depends on the current route depth (`/[locale]/notes` → `../charts` resolves to `/[locale]/charts`). If the notes page ever moves or gains nested routes, this breaks. The filing/document handlers use `setViewerDoc` (in-page), so they don't have this problem. Consider using the Next.js router or an absolute path like `` `/${locale}/charts` `` for robustness. |
| 4 | `notes/page.tsx:269` | **`handleNodeClick` dependency array is missing `getNote`.** Not introduced by this PR, but the new `chart` branch doesn't need deps since it only uses `window.open`. Still, the existing closure over `getNote` without listing it is a latent stale-closure risk. |

### 🔵 Nitpick

| # | File | Finding |
|---|------|---------|
| 1 | `tiptap-viewer.tsx:30-31` | The `chart` variant follows the `note` pattern (no `citationKey`), which is correct. Clean addition to the union. |

---

## Positive Feedback

- The change follows the established pattern exactly — new union variant, new selector entry, new case in the handler. Easy to review and low risk.
- The PR description and test plan are clear and thorough.
- Correctly identified that the `data-chart-link-id` attribute was already rendered but had no click handler wired up.

---

## Conclusion

**Request Changes** — the single substantive issue is that `clicked.chartId` is extracted but never used in the navigation URL (#1 above). The link just opens the charts listing page rather than navigating to the specific chart. This should be addressed before merging since it undermines the purpose of the PR (making chart links functional). Items #2–4 are optional improvements.

nvasilevv:
## Addressing review findings

**#1 (fixed)** — Chart clicks now deep-link to the specific chart via `?chartId=<id>`. Added `useSearchParams` to the charts page to auto-select the matching chart on load.

**#3 (fixed)** — Replaced the fragile relative `../charts` path with an absolute `/${locale}/charts` using `useParams` from Next.js.

**#2 (no action)** — Silent failure on missing `chartId` is consistent with how filing, document, and note links all behave today. Better addressed holistically if we ever add user-facing error toasts, not as a one-off for charts.

**#4 (no action)** — `getNote` is a module-level import from `@/lib/api` whose reference never changes between renders, so the missing dependency is harmless in practice. The exhaustive-deps lint rule would flag it, but it can't cause a stale closure. If `getNote` ever moves to a hook or prop, the linter will catch it at that point. Not worth a separate issue.

PR #50, opened 2026-04-16
Fix new chats not appearing dynamically in chat history (#28)

### LINKED ISSUE #28: Chats - new chats are not displayed dynamically in the list of hats
When a new chat is created it only appears in the list after a full page reload. It should automatically appear in the list after it has been created

nvasilevv:
## Summary
- Switch `chatListChanged` from a bare `Event` to a `CustomEvent` carrying the new chat's `id` and `title`
- Chat history panel now **optimistically prepends** the new chat entry instantly, then reconciles with the server via background re-fetch
- Auto-expands the history panel when a new chat is created so the user sees the entry immediately
- Backend now includes `title` in the `chat-metadata` SSE event
- Removes the loading flash on re-fetch (only shows loading spinner on initial load)

Closes #28

## Test plan
- [ ] Create a new chat from `/ask` by sending the first message — verify it appears in the sidebar history immediately
- [ ] Create a chat from a filing row ("Chat with filing" button) — verify it appears in history
- [ ] Create a chat from a company view ("Chat with company" button) — verify it appears in history  
- [ ] Verify the history panel auto-expands when a new chat is created
- [ ] Verify that after the optimistic update, the entry is replaced by the real data from the server (title matches)
- [ ] Verify deleting and renaming chats still work correctly


nvasilevv (commented):
## PR Review

### Summary
This PR switches the `chatListChanged` event from a bare `Event` to a `CustomEvent` carrying the new chat's `id` and `title`. The chat history sidebar now optimistically prepends new entries instantly (and auto-expands), then reconciles with a background re-fetch. The backend also includes `title` in the `chat-metadata` SSE event.

---

### Review Findings

#### 🔴 Critical/Security

**None found.** No injection vectors, no secrets exposed, no auth bypasses.

#### 🟡 Improvement

1. **Double `chatListChanged` dispatch for SSE-created chats (`ask-panel.tsx:509`)**
   When a user sends the first message in a new chat, `createInterceptingFetch` fires the `chatListChanged` event on line 509. But the `chat-history.tsx` handler calls `fetchChats()` unconditionally on *every* `chatListChanged` event. Meanwhile, the `company-view.tsx` and `feed-row.tsx` paths dispatch the event *and then* navigate to `/ask?chat=...`, which will trigger yet another SSE stream. This is fine functionally, but means **every new-chat creation triggers 2 API calls to `getChats()`** (one from the optimistic handler, one from the initial `useEffect` on mount after navigation). Consider debouncing `fetchChats()` or skipping the re-fetch when the optimistic update already covered it.

2. **`setLoading(false)` without a matching `setLoading(true)` (`chat-history.tsx:127`)**
   The `setLoading(true)` was removed from `fetchChats`, but `setLoading(false)` remains. After the initial mount (which does set `loading` to `true` via the default state), subsequent background re-fetches will call `setLoading(false)` on an already-false value — harmless but confusing. Consider either removing `setLoading(false)` from `fetchChats` entirely (and only managing it on mount), or keeping the symmetry with a conditional set.

3. **`hasattr(chat_row, "title")` guard is unnecessary (`routes.py:2653`)**
   `chat_row` is always a `Chat` model instance (which always has `title`). The `hasattr` check adds defensive noise for a case that can't happen. `if chat_row and chat_row.title:` would be sufficient.

4. **Potential race: optimistic entry could persist if `fetchChats()` fails**
   If the background `fetchChats()` call fails (network blip, expired token), the optimistic entry with a potentially stale title stays in the list permanently until next manual refresh. Consider adding error handling or a retry.

#### 🔵 Nitpick

1. **Type assertion on `CustomEvent` detail (`chat-history.tsx:139-141`)**
   The `(e as CustomEvent).detail as { id: string; title: string } | undefined` cast could be replaced with a typed helper or a `CustomEvent<{ id: string; title: string }>` generic for slightly better type safety.

2. **Magic string `"Untitled"` (`chat-history.tsx:148`)**
   This fallback title is used in one place. If the backend ever changes its default title, they could drift. Minor, but worth noting.

---

### Positive Feedback

- **Optimistic UI pattern is well-implemented.** The `setChats` updater correctly checks for duplicates (`prev.some(c => c.id === detail.id)`) before prepending, preventing double entries.
- **Auto-expand on new chat** is a nice UX touch — users immediately see their new chat appear.
- **Backend change is minimal and safe** — conditionally adding `title` to an existing SSE event is backwards-compatible.
- **Clean separation of concerns** — each entry point (company view, feed row, ask panel) dispatches the same event shape, and the history component handles it in one place.
- **The PR description and test plan are thorough.**

---

### Conclusion

**Approve** — with minor suggestions. The core logic is correct, the optimistic update pattern is clean, and there are no security concerns. The improvement items above are optional quality-of-life enhancements, none are blocking.

nvasilevv:
## Review — Fix new chats not appearing dynamically in chat history

### Summary
Switches `chatListChanged` to a `CustomEvent` carrying `{id, title}`, optimistically prepends the new chat in `ChatHistory`, and reconciles via a debounced background re-fetch. Backend now includes `title` in the `chat-metadata` SSE frame, and `getChats()` returns `null` on failure so optimistic entries aren't clobbered. Scoped, low-risk change.

### Findings

🔴 **Critical / Security** — none.

🟡 **Improvements**
- `frontend/src/components/layout/chat-history.tsx:129` — `fetchTimerRef` is never cleared on unmount. If `ChatHistory` unmounts within the 200 ms debounce window, the timer still fires and calls `setChats` on an unmounted component. Add `clearTimeout(fetchTimerRef.current)` in the cleanup of the listener effect (or a dedicated unmount effect).
- `chat-history.tsx:160` — `setExpanded(true)` fires on *every* `chatListChanged`, including deletes/renames if this event is ever reused, and overrides a user who just collapsed the panel. Consider only auto-expanding when the incoming `detail.id` is genuinely new (the `prev.some(...)` branch already tells you that).
- `chat-history.tsx:134` — on a `null` result we silently keep stale state. Fine for transient failures, but no retry/backoff means a user who loses auth mid-session sees a frozen list forever. A single retry or surfacing the error state would be more robust (non-blocking).
- `ask-panel.tsx:508` — optimistic entry is dispatched with `title ?? ""`, which renders as "Untitled" until the debounced refetch. Expected, but worth noting: if the backend title update races the 200 ms refetch, the user briefly sees "Untitled" then the real title. Consider a second refetch after the first AI response completes for snappier title population.
- `backend/app/api/routes.py:2654` — the `if chat_row and chat_row.title` guard is fine, but consider always including `title` (empty string when absent) to keep the SSE schema stable; optional fields on structured events are easy to forget on the consumer side.

🔵 **Nitpicks**
- `UNTITLED_CHAT` constant is good; consider sharing it with any other "Untitled" fallbacks elsewhere.
- `useRef<ReturnType<typeof setTimeout>>(undefined)` — the explicit `undefined` argument is fine but `useRef<ReturnType<typeof setTimeout> | undefined>(undefined)` is the conventional form.
- The `chatListChanged` event payload shape (`{id, title}`) is duplicated across three dispatch sites — a tiny `dispatchChatListChanged(id, title)` helper in `lib/` would keep the contract in one place.

### Positive
- Good reasoning in the comments (`null`-vs-`[]` on failure, debounce rationale, stall-timer in `createInterceptingFetch`).
- Dedup-on-id in the optimistic insert (`prev.some((c) => c.id === detail.id)`) correctly handles the race between the optimistic dispatch and the debounced refetch.
- Backend change is minimal and backwards-compatible (new optional field).

### Conclusion
**Approve with minor changes** — @nvasilevv could you address the 🟡 items above (especially the `fetchTimerRef` cleanup and the unconditional `setExpanded(true)`) before merging? The 🔵 nitpicks are optional polish. Thanks!

nvasilevv:
## Review Summary

Tight, well-scoped fix for #28. The earlier review rounds (timer cleanup, conditional expand, stable SSE schema) landed cleanly. Approving, but flagging a few non-blocking items worth addressing in a follow-up — please take a look:

### 🟡 Please address
- **`chat-history.tsx:129` — initial-load error path.** `fetchChats().then(() => setLoading(false))` clears the spinner even when `getChats()` returns `null` (network flake / token race), leaving an empty panel with no retry. Add a one-time retry or surface an error state; at minimum keep `loading=true` until a successful fetch once.
- **`chat-history.tsx:145` — `isNew` captured across the `setChats` updater.** Under strict-mode dev double-invoke the updater fires twice and writes `isNew = true` twice — harmless today, fragile tomorrow. Safer: compute `isNew` outside the updater (via a `chatsRef` kept in sync), or drive the expand from a `useEffect` reacting to length change.
- **`routes.py:2654` — `chat_row` nullability.** At this point `chat_row` should exist (just committed); silently coercing `None` to `""` could mask a real bug. Add an assertion or a warning log if `chat_row is None`.
- **Double dispatch on the `/ask` flow.** `feed-row` / `company-view` dispatch `chatListChanged` before `router.push`, and `ask-panel` dispatches it again on `chat-metadata`. The de-dup + debounce makes it correct, but it's non-obvious — either add a one-line comment explaining the intentional double-fire, or drop the pre-navigation dispatch and rely on `ask-panel` emitting with the real title.

### 🔵 Nits (optional)
- `useRef<ReturnType<typeof setTimeout>>(undefined)` → prefer `useRef<ReturnType<typeof setTimeout> | undefined>(undefined)` for explicit nullability.
- Consider `title: chat_row.title or None` on the SSE payload and let the frontend fall back, rather than coercing empty-string on both sides.
- Optimistic entry uses `updated_at = now`, which pins it to the top even if the server later returns an older timestamp (shouldn't happen, but worth noting).

### Positive
- The `ChatListItem[] | null` signal from `getChats` is the right call — cleanly distinguishes "failed, preserve optimistic" from "empty".
- Debounced re-fetch correctly collapses the optimistic + navigation-mount + metadata event storm into one round-trip.
- `isNew`-gated auto-expand respects a user who just collapsed the panel.

**Verdict: Approve** — please address the 🟡 items in a follow-up when you get a chance.

nvasilevv:
## Summary

This PR fixes new chats not appearing dynamically in the sidebar. It renames the browser event from `chat-list-changed` → `chatListChanged` with a `CustomEvent` payload (id + title), optimistically prepends the new chat on dispatch, and debounces a background re-fetch to reconcile with the server. Backend additionally includes `title` in the `chat-metadata` SSE event. Overall a solid, targeted fix.

## Review Findings

### 🟡 Improvement
- **Optimistic title never updated once the SSE delivers the real one.** In `chat-history.tsx:186–198`, the handler adds an entry only when the id is absent (`!chatsRef.current.some(...)`). The second dispatch from `ask-panel.tsx:512` (carrying the real `title` from the SSE `chat-metadata` event) is thus effectively ignored — the placeholder stays "Untitled" until the debounced re-fetch (≥200 ms) completes. If you want the SSE-delivered title to land instantly, update the entry when `detail.title` is non-empty and differs. Today the refetch papers over it, but the comment on line 509 ("real title from the SSE… de-dupes by id, so the double-fire is intentional") promises behaviour the code doesn't deliver.
- **Redundant existence check.** Same handler checks `chatsRef.current.some(...)` *and* the functional `setChats((prev) => prev.some(...) ? prev : [...])` — the ref check is now just an optimisation to skip the setState + ISO string allocation, not for correctness. Either drop the outer check or drop the inner guard; keeping both is confusing and invites drift.
- **`feed-row` / `company-view` dispatch title from `createChatFromFiling`/`createChatFromCompany` response.** Worth verifying those endpoints return the generated title (not an empty default) — otherwise the optimistic entry briefly shows "Untitled" even though we had a real title available. If the API returns `""`, prefer waiting for the SSE event (the `ask-panel` path) rather than dispatching pre-navigation.
- **Debounced fetch not cancelled on initial-load path.** The init effect tracks `cancelled`, but `debouncedFetchChats` scheduled from the event handler can still resolve after unmount and call `setChats` on an unmounted component. Modern React tolerates this quietly, but passing an `AbortController` through `getChats`, or clearing the debounce in the init cleanup, would be cleaner.

### 🔵 Nitpick
- **PR description mentions auto-expand**, but the diff contains no expand logic. Either it landed in a prior commit on the branch (FYI) or got dropped — worth confirming the test plan item "Verify the history panel auto-expands" still passes.
- **`_meta: dict[str, Any]`** — the explicit annotation adds little since `_sse` already accepts a dict; the `: Any` import was already present so no new dep. Minor.
- **Event name rename is a micro-breaking change.** Confirmed via grep there are no remaining `chat-list-changed` listeners, so this is fine — just noting for anyone rebasing a long-lived branch.
- **`(chat_row.title if chat_row else "") or ""`** — the trailing `or ""` is redundant since the ternary already yields `""` when `chat_row` is None; collapse to `chat_row.title if chat_row else ""`.

## Positive Feedback
- The `chatsRef` pattern with a one-shot sync effect is the right call — avoids the stale-closure trap without re-subscribing the listener on every state change, and the inline comment explains why.
- Nice touch returning `null` from `getChats` on failure and using that to *preserve* the optimistic entry rather than clobbering — the prior `return []` would have wiped the just-added chat on any transient 401/5xx.
- Debouncing the refetch to 200 ms to collapse the optimistic-dispatch + navigation-mount + metadata-event storm into a single round trip is a clean solution to a real thundering-herd problem.
- Good defensive log for the `chat_row is None` branch in `routes.py` instead of silently emitting an empty title.
- The two-attempt initial load with `cancelled` flag handles the token-race cleanly.

## Conclusion
**Comment** (close to Approve). No blockers. The optimistic-title-never-updates point is worth addressing since the code comment asserts behaviour the handler doesn't implement, but the debounced refetch makes it a UX polish rather than a correctness issue. The redundant existence check and duplicate `or ""` are trivial cleanups.

nvasilevv:
Re-reviewed the latest changes — all material comments addressed:

**Addressed ✅**
- Optimistic title now upserts when a later dispatch carries a non-empty, different title (`chat-history.tsx:160-162`).
- Redundant existence check removed; single guard lives inside the `setChats` updater.
- Redundant `or ""` dropped in `routes.py` — now `chat_row.title if chat_row else ""`.
- Timer cleanup on unmount added.

**Partially addressed**
- Debounced-fetch-after-unmount: `clearTimeout` cancels scheduled calls, but a fetch already in-flight when the component unmounts can still call `setChats`. Harmless in practice.

**Not addressed (acceptable)**
- `feed-row` / `company-view` still trust the API's `chat.title` — fine assuming those endpoints always return a real title.
- `_meta: dict[str, Any]` annotation kept — stylistic only.
- Auto-expand logic isn't in this diff but appears to have landed in commit `67beecf`.

LGTM 👍

PR #52, opened 2026-04-16
Hook charts up to datasets for user-friendly chart creation

### LINKED ISSUE #51: Charts - not really usable
Charts are not really usable in the app currently. From the UI, users should pass some JSON array, where are they supposed to get this array from? It doesn't make much sense to be like that. It should be somewhat more user-friendly.

I believe we should hook it up to the datasets. If a user has a dataset (one of the datasets available in the Datasets tab), it should be able for the user to construct a chart in addition to that, by just creating a new chart and then, somewhere in between choosing the type of chart and then underneath the chart, maybe the user can select the dataset.

nvasilevv:
## Summary
- Replace raw JSON textarea in chart editor with a **3-tab data interface**: "From Dataset", "Data Source", and "Manual JSON"
- Users can now select any dataset from a dropdown, preview its data in a table, and auto-populate a chart — no JSON editing required
- X-axis key is now a dropdown of detected keys instead of a free-text input
- Backend: added `dataset` as a new populate source type that fetches DatasetRows with Filing+Issuer metadata

Closes #51

## Test plan
- [ ] Create a new chart → verify "From Dataset" tab is shown by default
- [ ] Select a dataset with data → verify data preview table appears and chart renders
- [ ] Change x-axis key dropdown → verify chart updates
- [ ] Switch to "Data Source" tab → verify eod_prices/xbrl_facts/analyst_estimates still work
- [ ] Switch to "Manual JSON" tab → verify textarea works as before
- [ ] Save chart from dataset → verify data_source metadata is stored
- [ ] Edit existing chart → verify correct tab/state is restored


nvasilevv:
## PR #52 Review: Hook charts up to datasets for user-friendly chart creation

### Summary

This PR replaces the raw JSON textarea in the chart editor with a 3-tab data interface ("From Dataset", "Data Source", "Manual JSON"), adds a new `dataset` populate source type on the backend, and introduces an XBRL concept search endpoint with ticker scoping. The goal is to let users create charts from datasets without manually editing JSON.

---

### Review Findings

#### 🔴 Critical / Security

| # | Finding | Location |
|---|---------|----------|
| 1 | **No authorization on XBRL concept search endpoint.** The `email` param is injected via `Depends(get_current_user_email)` but never used — any authenticated user can search any ticker's XBRL concepts. This is likely fine for read-only concept metadata, but the unused dependency is misleading. If there's org-scoped data isolation intent, this is a gap. | `backend/app/api/charts.py:23-33` |
| 2 | **Dataset access check uses 404 for authz failure** (`detail="Dataset not found"`). This is fine from an information-leakage standpoint (doesn't reveal existence), but consider returning 403 for audit logging so you can distinguish "not found" from "forbidden" internally. | `backend/app/api/charts.py:109-110` |
| 3 | **Unbounded dataset row fetch on frontend.** `getDatasetData(datasetId, { page: 1, page_size: 500 })` loads up to 500 rows into browser memory, then `JSON.stringify`s them into `dataJson` state. For large datasets this could cause UI freezes. The backend `populate_chart_data` has a `limit` param but the frontend dataset path bypasses it entirely. | `frontend/src/app/[locale]/charts/page.tsx:281` |

#### 🟡 Improvement

| # | Finding | Location |
|---|---------|----------|
| 4 | **Missing error handling in `loadDatasetData`.** If `getDatasetData` throws (network error, 500), the function will crash without setting `datasetDataLoading` back to `false`, leaving the UI stuck on "Loading dataset data..." forever. Wrap in try/catch/finally. | `frontend/src/app/[locale]/charts/page.tsx:278-309` |
| 5 | **Same issue in `searchConcepts`** — no try/catch around `searchXbrlConcepts`. If the call throws, `xbrlSearching` stays `true` permanently. | `frontend/src/app/[locale]/charts/page.tsx:312-317` |
| 6 | **`loadDatasets` silently swallows errors.** If `getDatasets()` fails, `datasets` stays `[]` and the user sees "No datasets available" with no indication something went wrong. | `frontend/src/app/[locale]/charts/page.tsx:80-84` |
| 7 | **XBRL populate lost the ticker requirement.** The old code only sent `ticker` for `eod_prices`. The new code sends it for both `eod_prices` and `xbrl_facts`, which is correct — but the backend's `query_xbrl_facts_by_ticker` receives `payload.ticker or None`. If the user doesn't enter a ticker for XBRL, the backend silently gets `None` and may return unscoped global results. Consider validating that `ticker` is required when `source_type == "xbrl_facts"`. | `backend/app/api/charts.py:69-73` |
| 8 | **`data_source` field set on save even for new charts.** When `dataMode === "dataset"`, the payload sends `data_source: "dataset:<uuid>"`. But `data_source` isn't in the `CreateChartRequest` schema (only in update). Verify the create endpoint accepts this field; if not, it's silently dropped. | `frontend/src/app/[locale]/charts/page.tsx:349-351` |
| 9 | **`ticker: payload.ticker or None` is a Python falsy trap.** If `payload.ticker` is an empty string `""`, `"" or None` evaluates to `None`, which is the desired behavior. But it's fragile — consider explicit `payload.ticker if payload.ticker else None` or validating at the Pydantic level with `min_length=1`. | `backend/app/api/charts.py:70` |
| 10 | **Inner join on Filing/Issuer** means dataset rows without a valid `filing_id` or `issuer_id` are silently dropped. If datasets can have rows with null FK references, this will return fewer rows than expected with no error. | `backend/app/api/charts.py:113-116` |

#### 🔵 Nitpick

| # | Finding | Location |
|---|---------|----------|
| 11 | **Inline import inside endpoint handler.** `from app.services.repository import list_xbrl_concepts_for_ticker` is inside the function body. This is consistent with the existing `populate_chart_data` pattern in this file, so it's not a regression — but it's worth noting these could be top-level imports. | `backend/app/api/charts.py:31` |
| 12 | **False-positive "no concepts found" message.** The condition `xbrlConcepts.length === 0 && !xbrlSearching && dsTicker && xbrlKeyword` shows "No concepts found" even before the user has ever searched — as long as they've typed in both fields. Should also track whether a search has been performed. | `frontend/src/app/[locale]/charts/page.tsx:696` |
| 13 | **`allKeys.slice(0, 8)` magic number.** The 8-column limit for the preview table is reasonable but undocumented. A named constant would be clearer. | `frontend/src/app/[locale]/charts/page.tsx:580` |

---

### Positive Feedback

- The **3-tab data interface** is a well-designed UX pattern — it cleanly separates the three data entry workflows without overloading any single view.
- The **dataset access check** (owner, global, or same org) is correctly implemented and avoids leaking dataset existence via 404.
- **XBRL concept search with semantic aliases** is a nice touch — it makes the data source discovery much more user-friendly compared to requiring users to know exact concept names.
- The **data preview table** with x-axis highlighting is a good addition for helping users understand what they're charting before saving.
- Good refactoring of the XBRL populate path to use `query_xbrl_facts_by_ticker` instead of raw SQL in the endpoint.

---

### Conclusion

**Request Changes** — The main concerns are:

1. Add try/catch/finally to `loadDatasetData` and `searchConcepts` to prevent stuck loading states (items 4, 5)
2. Consider adding backend validation that `ticker` is required for `xbrl_facts` source type (item 7)
3. Evaluate whether 500-row unbounded client-side fetch is acceptable or needs pagination/streaming (item 3)

Items 4 and 5 are the most likely to affect users in production. The rest are improvements worth considering but not blocking.

nvasilevv:
## Follow-up Review (commit cf7d15fe)

The latest commit addresses the key concerns well. Here's the scorecard:

| # | Finding | Status |
|---|---------|--------|
| 4 | **`loadDatasetData` missing error handling** | ✅ Fixed — wrapped in try/catch/finally, loading state always resets |
| 5 | **`searchConcepts` missing error handling** | ✅ Fixed — same pattern, try/catch/finally |
| 6 | **`loadDatasets` silently swallows errors** | ✅ Fixed — now shows "Failed to load datasets." error message |
| 7 | **XBRL ticker not validated on backend** | ✅ Fixed — explicit `ticker` required check + `.strip()` on both ticker and concept |
| 9 | **`payload.ticker or None` falsy trap** | ✅ Fixed — now uses `payload.ticker.strip()` after the validation gate |
| 12 | **False-positive "no concepts found"** | ✅ Fixed — added `xbrlHasSearched` flag, resets on ticker change |

**Still open (lower severity):**

| # | Finding | Status |
|---|---------|--------|
| 1 | Unused `email` dependency on XBRL endpoint | Not addressed (minor) |
| 2 | 404 vs 403 for authz on dataset access | Not addressed (minor, current approach is fine) |
| 3 | 500-row unbounded client fetch | Not addressed — still worth considering for large datasets |
| 8 | `data_source` field on chart create | Not addressed — worth verifying |
| 10 | Inner join silently drops rows with null FKs | Not addressed |

The critical items (4, 5, 7) are all fixed. The remaining items are lower severity. **Good to merge** — the open items are either minor or edge cases that can be addressed later.

PR #58, opened 2026-04-16
Replace notes sidebar toggle with full-height collapse rail

### LINKED ISSUE #57: Style of left note pane that is collapsible
So currently, the style of this collapsible pane is using this small icon which one can click on and choose Show notes list. It opens up all notes, and then one can click on it again to hide the notes list, and it hides the notes list. Once the notes are hidden, this icon is actually right on top of the name of the note that I'm currently looking at, so it's blocking part of the characters I was thinking of.

Can we have something that's a bit of a different thing, not the separate icon in the collapsed state, but maybe something that's a bit of a longer thing that goes along the whole height of the window? There is a single arrow in the middle which I can click on to open up, and I can click on the same thing, but it is pointing in the other direction to close this list of notes. Will be more elegant, I believe. Can we use it also everywhere we do collapsing things, but not for the main left pane where it is SeeWire live heat and so on? Let's keep that one changed, but everywhere else where we use these collapsible panes on the left side to list where listing stuff is, let's have them in this format I described.

nvasilevv:
## Summary
- Replaces the small `PanelLeftClose`/`PanelLeftOpen` icon buttons on the notes sidebar with a narrow full-height vertical rail between the sidebar and content area
- The rail has a centered chevron arrow (`‹` when expanded, `›` when collapsed) and spans the entire window height, making it a much larger and more intuitive click target
- The rail is always visible regardless of sidebar state — no more floating button overlapping the note title

Closes #57

## Test plan
- [ ] Open notes page, verify the vertical rail appears between sidebar and content
- [ ] Click the rail to collapse the sidebar — arrow should flip to point right
- [ ] Click the rail again to expand — arrow should flip back to point left
- [ ] Verify the rail hover state highlights correctly
- [ ] Test on mobile viewport — sidebar should still default to collapsed


nvasilevv (commented):
## PR Review: Replace notes sidebar toggle with full-height collapse rail

### Summary

Replaces the small `PanelLeftClose`/`PanelLeftOpen` icon buttons with a narrow full-height vertical rail positioned between the sidebar and content area. The rail uses chevron arrows that flip direction based on state and is always visible, improving discoverability and click-target size. Net reduction of 3 lines — clean simplification.

**Approve** ✓ — the core approach is solid. A couple of minor items worth addressing below.

---

### Findings

**🟡 Potential double border when sidebar is expanded**

The sidebar div retains `border-r border-border`, and the new rail also has `border-r border-border`. When the sidebar is open the layout is: `sidebar (border-r) | rail (border-r) | content` — the sidebar's right border sits flush against the rail's left edge, which may create a visual double-line effect. Consider removing `border-r` from the sidebar div since the rail now owns that boundary.

**🟡 Touch target width on mobile**

The rail is `w-3.5` (14px). While its full height gives it a large total area, 14px is below the recommended 44px minimum touch target width (WCAG / Apple HIG). On mobile viewports where the sidebar defaults to collapsed, the rail is the only way to re-open it. Consider bumping to `w-5` or adding invisible padding to widen the effective tap area.

**🔵 Accessibility nitpick**

Adding `aria-label={sidebarOpen ? "Collapse sidebar" : "Expand sidebar"}` alongside the existing `title` would improve screen-reader support — the chevron icon alone doesn't convey meaning.

---

### What's good

- Functional updater `setSidebarOpen((o) => !o)` avoids stale closure issues.
- Consolidating two conditional buttons into one always-visible rail is a clean simplification — less conditional rendering, fewer z-index concerns.
- Net negative diff with no dead imports left behind.

nvasilevv (commented):
All three review items addressed cleanly in ec219c3:

1. **Double border** — `border-r` removed from the sidebar div; the rail now owns the only vertical border. ✓
2. **Touch target** — Rail widened from `w-3.5` (14px) to `w-5` (20px). ✓
3. **Accessibility** — `aria-label` added with state-aware text. ✓

LGTM — good to merge.

PR #61, opened 2026-04-16
Move chat history from sidebar into collapsible pane on /ask page

### LINKED ISSUE #60: Chat lists - consolidate into how rest are working
So chat lists are the only lists currently that are displayed in the left pane. Everything else is clickable functionality, and if it contains a list, it is listed as a separate pane next to the main pane, which is collapsible. This is the approach for chats. This is the approach for research notes. I want us to move to the same approach for chats as well.

nvasilevv:
## Summary
- Removes chat history from the global navigation sidebar and moves it into a dedicated collapsible left pane on the `/ask` page, matching the pattern used by Research Notes and other list views
- Chat history pane features a header with count badge and "New Chat" button, always-visible search, full-height scrollable list with active chat highlighting, and a thin collapse/expand rail (ChevronLeft/ChevronRight)
- Sidebar state persists via localStorage and defaults to closed on mobile (<768px) with auto-close on chat selection

Closes #60

## Test plan
- [ ] Navigate to `/ask` — chat history pane appears on the left with collapse rail
- [ ] Toggle the collapse rail — pane hides/shows, state persists across page refresh
- [ ] Click a chat — loads in main pane, highlighted in list
- [ ] Create a new chat (send a message) — list auto-refreshes with the new chat
- [ ] Rename, delete, and pop-out actions still work from the chat list
- [ ] Search filtering works correctly
- [ ] On mobile viewport, sidebar starts closed and auto-closes after selecting a chat
- [ ] `/ask/popout` page is unaffected
- [ ] Main navigation sidebar no longer shows chat history section


nvasilevv:
## PR #61 Review: Move chat history from sidebar into collapsible pane on /ask page

### 1. Summary

This PR extracts the chat history from the global navigation sidebar and relocates it to a dedicated collapsible left pane on the `/ask` page. The `ChatHistory` component is refactored from an accordion-style widget into a full-height panel with header, always-visible search, active chat highlighting, and a thin collapse/expand rail. Sidebar state persists via `localStorage` with sensible mobile defaults.

---

### 2. Review Findings

#### 🔴 Critical/Security

- **None found.** No injection vectors, no secrets exposed, no auth/authz changes, no new dependencies.

#### 🟡 Improvement

| # | File | Finding |
|---|------|---------|
| 1 | `ask/page.tsx:14-19` | **SSR hydration mismatch risk.** The `useState` initializer reads `window.matchMedia` and `localStorage`, which don't exist on the server. Because `"use client"` components still SSR on first render in Next.js App Router, the server will always compute `true` while the client may compute `false` (mobile) — React will log a hydration warning. Consider initializing to a stable default (e.g. `true`) and running the `localStorage`/media-query logic in a `useEffect` on mount instead. |
| 2 | `ask/page.tsx:26-30` | **Mobile breakpoint check is a point-in-time snapshot.** `handleMobileSelect` calls `window.matchMedia` on every select but never listens for viewport changes. If a user resizes their browser across the 768px boundary, the behaviour won't adapt. A minor concern — consider a `useMediaQuery` hook if this pattern spreads. |
| 3 | `chat-history.tsx` (diff) | **`ChatHistory` now renders a fragment (`<>...</>`) instead of a single wrapper div.** The parent in `page.tsx` wraps it in `<div className="flex w-64 ...">`, so layout is fine. However, the component no longer owns its own width — if reused elsewhere without the wrapper, it would break. The tight coupling is acceptable for now but worth noting. |
| 4 | `ask-panel.tsx:509` | **`CustomEvent("chat-list-changed")` is a global window event.** This works but is an implicit contract between `AskPanel` and `ChatHistory`. If either component is ever unmounted/remounted independently, the listener could be stale. A shared context or callback prop would be more explicit, but this is fine for the current scope. |
| 5 | `chat-history.tsx` | **`Loader2` is imported but `ChevronRight` import was removed.** Good cleanup — just confirming the unused import (`ChevronRight`) was correctly removed since the expand/collapse chevron now lives in `page.tsx`. Verified: no remaining references. |

#### 🔵 Nitpick

| # | File | Finding |
|---|------|---------|
| 1 | `ask/page.tsx:39` | The sidebar div uses a fixed `w-64` (256px). Consider extracting this to a constant or CSS variable if it's referenced elsewhere or may be adjusted later. |
| 2 | `chat-history.tsx` | The `cn` utility import was added — good, it aligns with the existing pattern used across the codebase for conditional classNames. |
| 3 | `sidebar.tsx` | The removal is clean — the `ChatHistory` import and its container div are gone, replaced by a simple spacer. No orphaned code left behind. |

---

### 3. Positive Feedback

- **Clean architectural separation.** Moving chat history from the global sidebar into a page-specific pane is the right call — it reduces sidebar complexity and gives the `/ask` page ownership of its own state. Well-scoped change.
- **Good mobile UX defaults.** Auto-closing the sidebar on mobile and persisting state via `localStorage` is thoughtful. The `handleMobileSelect` auto-close on chat selection is a nice touch.
- **`CustomEvent` for cross-component refresh** is a pragmatic, low-overhead solution for notifying the chat list when a new chat is created, avoiding prop drilling through deeply nested components.
- **The collapse rail pattern** is consistent with other parts of the app (Research Notes, per the PR description), maintaining UI consistency.
- **Clean diff** — the old accordion/expand logic is fully removed, not left behind as dead code.

---

### 4. Conclusion

**Comment** (leaning Approve)

The only actionable item is **#1 (hydration mismatch)** — the `useState` initializer accesses browser APIs that aren't available during SSR, which will produce React hydration warnings. This should be addressed before merge. The remaining items are minor suggestions. Otherwise, this is a clean, well-structured PR.

PR #62, opened 2026-04-16
Fix table rendering in document side viewer (#48)

### LINKED ISSUE #48: Document side viewer in chat and research notes not parsing tables properly?
Is it possible that the side viewer we have on the right side pane, when the user clicks on a citation by the agent or on a citation included in a research note, is not properly parsing the markdown tables? Is it possible?

nvasilevv:
## Summary

Filing documents opened in the side pane (via citation click in chat or research notes) were losing table formatting. This PR fixes table parsing and adds horizontal scroll support for wide tables.

### Problem

The document side viewer uses TipTap + `tiptap-markdown` to render markdown content. When `markdown-it` converts GFM pipe tables to HTML, it wraps rows in `<thead>` and `<tbody>` elements. ProseMirror's DOM parser does not list these as recognized block-level wrappers (they're absent from its internal `blockTags` set), so table rows inside them could be missed during document construction — causing tables to lose their structure.

Additionally, wide financial tables (common in filing documents) had no horizontal scroll, overflowing the side panel.

### Changes

- **`frontend/src/lib/tiptap/table-markdown.ts`** — Added a `parse.updateDOM` hook that flattens `<thead>`, `<tbody>`, and `<tfoot>` wrappers before ProseMirror processes the DOM. This ensures `<tr>` elements sit directly inside `<table>`, where ProseMirror reliably picks them up.

- **`frontend/src/app/globals.css`** — Added `overflow-x: auto` on TipTap's `.tableWrapper` div (the wrapper TipTap's `TableView` creates around every table). Also zeroed out table margins when inside `.tableWrapper` to prevent doubling with the wrapper's own margins.

### What's NOT affected

- **Chat answer tables** — These use ReactMarkdown + remark-gfm (completely separate rendering path). Unaffected.
- **Research note tables** — These already worked because content enters TipTap through the editor (not via markdown-it parsing). The `updateDOM` hook is additive and doesn't change behavior for content that already parsed correctly.

### Fix target
- [ ] Open /ask, ask about a company with financial filings, click a citation → verify tables render with proper headers, rows, and alignment in the document side viewer
- [ ] Verify wide tables scroll horizontally in the side pane

### Regression checks
- [ ] Open a research note containing a table → verify it still renders correctly
- [ ] Chat answers with tables → verify no regression (should look the same as before)

### No negative effects expected
- The `updateDOM` hook only flattens wrapper elements (`<thead>`/`<tbody>`/`<tfoot>`) that ProseMirror doesn't recognize anyway — content that already parsed correctly is unaffected
- CSS changes are scoped to TipTap's `.tableWrapper` class, which ReactMarkdown tables don't use

Closes #48


nvasilevv:
## Code Review

**Summary:** A small, focused fix (26 additions, 1 deletion) across two files. It resolves a ProseMirror table-parsing bug by flattening `<thead>`/`<tbody>`/`<tfoot>` wrappers before DOM ingestion, and adds `overflow-x: auto` on TipTap's `.tableWrapper` for wide financial tables. The root cause diagnosis is accurate and the fix is idiomatic.

---

### Findings

**🔵 Nitpick — Bundled unrelated CSS fix**
The link-selector expansion (`.filing-link`, `.document-link`, `.chart-link` added to `globals.css:229–232`) is unrelated to table rendering. Ideally belongs in its own commit for a clean history, but it's additive and harmless.

**🔵 Nitpick — Non-null assertion `wrapper.parentNode!`**
At `table-markdown.ts:149`, the `!` is safe since `wrapper` was just selected as a direct child of `table`. An explicit guard (`if (!wrapper.parentNode) continue`) would be more defensive, though there's no real risk here.

**🔵 Nitpick — `Array.from()` wrapping looks redundant without context**
Wrapping `querySelectorAll` in `Array.from()` before the `for...of` loop is the correct call (the live NodeList mutates during DOM manipulation), but a short inline comment would help the next reader understand why it's there.

---

### Positive Feedback

- **Surgical fix** — the `updateDOM` hook fires once at parse time, not on every render, so no performance cost.
- **`:scope >` selector** correctly limits flattening to direct children of each `<table>`, preventing mutation of nested tables.
- **Iteration order** is safe — `querySelectorAll("table")` is depth-first, so inner tables are flattened before outer wrappers are processed.
- **CSS margin zeroing** (`.tableWrapper > table { margin: 0 }`) cleanly prevents double-margin stacking without touching other table styles.
- The PR description is excellent — clearly explains which rendering paths are affected and which are not.

---

### Conclusion

**Approve** — the fix is correct, minimal, and well-reasoned. Please take a look at the three nitpicks above and decide whether any warrant a follow-up; none are blocking.

nvasilevv:
Thanks for the thorough review!

**Nitpick 1 (bundled link CSS):** Agreed — that belongs in its own commit. It slipped in since I was already in `globals.css`. Will keep those separate going forward.

**Nitpick 2 (non-null assertion):** Fixed — replaced `wrapper.parentNode!` with an explicit `if (!wrapper.parentNode) continue` guard.

**Nitpick 3 (`Array.from` reasoning):** Added a short inline comment explaining that the snapshot is required because the NodeList is live and mutates as children are moved out.

Pushed in d752e4a.

nvasilevv:
## Code Review

**Summary:** Tight 26-line, two-file fix for table rendering in the document side viewer. The root cause analysis (ProseMirror's DOMParser ignoring `<thead>`/`<tbody>`/`<tfoot>` wrappers emitted by `markdown-it`'s table tokenizer) is correct, and the chosen mitigation — a `parse.updateDOM` hook on the custom `TableWithMarkdown` extension — is the right escape hatch. The accompanying CSS adds horizontal scroll on `.tableWrapper` for wide financial tables.

---

### Findings

**🟡 Improvement — `Array.from()` snapshot comment is factually incorrect**
At `frontend/src/lib/tiptap/table-markdown.ts:142–145`, the comment claims the snapshot is required because "the NodeList is live and mutates as we move children out." This is wrong: `Element.querySelectorAll()` returns a **static** `NodeList` per the DOM spec — it does not reflect later DOM mutations. (Live collections come from `getElementsByTagName` / `getElementsByClassName`.) The `Array.from(...)` call here is harmless but strictly unnecessary; either drop the wrapper, or — if you keep it for symmetry with upstream `tiptap-markdown` patterns (e.g. `task-list.js` uses the same `[...nodeList]` spread) — fix the comment to say *"upstream tiptap-markdown extensions wrap querySelectorAll the same way; following the convention"* rather than asserting a liveness property that isn't true.

**🔵 Nitpick — `:scope >` ordering with nested tables**
`querySelectorAll("table")` walks in document order (outer-first). The `:scope >` constraint correctly limits each pass to the current table's direct children, so nested tables are handled correctly when iteration reaches them later. Worth a one-liner comment to reassure future readers, since the safety isn't obvious at a glance.

**🔵 Nitpick — `parse:` block formerly contained only a comment**
Before this PR, `parse: {}` held a single explanatory comment. Now that real behavior lives there, the docstring at the top of the file (lines 1–13) only mentions the *serializer* round-trip. Consider extending the file-level JSDoc to mention that the extension also patches markdown-it's table HTML to satisfy ProseMirror's parser — it'll save the next reader a `git blame`.

---

### Verified Claims

- ✅ **`parse.updateDOM` is a real, supported hook.** Confirmed in `tiptap-markdown/src/parse/MarkdownParser.js:34` — it iterates extensions and calls `getMarkdownSpec(extension)?.parse?.updateDOM?.call(...)` exactly once per parse, so there's no per-render cost.
- ✅ **Pattern matches upstream conventions.** `tiptap-markdown`'s own `task-list.js`, `task-item.js`, and `code-block.js` use the same `parse.updateDOM` shape.
- ✅ **CSS scoping claim holds.** `.tableWrapper` is created at runtime by `@tiptap/extension-table`'s `TableView`; `ReactMarkdown` (used in `ask-panel.tsx:241–247`) does not produce that class, so chat-answer tables are unaffected by the new rules.
- ✅ **No double-margin issue.** `.prose-ceewire .tableWrapper > table { margin: 0 }` correctly zeroes the inner table margin to compensate for the new wrapper margins.
- ✅ **Previous review feedback addressed in d752e4a** (parentNode guard added; Array.from comment added — though see finding above re: that comment's accuracy).

---

### Positive Feedback

- Surgical, low-blast-radius fix. The hook only mutates DOM nodes that ProseMirror was going to drop anyway.
- `:scope >` was the right choice — bare `thead, tbody, tfoot` would have flattened nested tables' wrappers when processing the outer table.
- The PR description is genuinely useful: it spells out which rendering paths are affected (TipTap markdown ingest) and which aren't (ReactMarkdown chat answers, TipTap editor input), which is exactly what a reviewer needs to scope the regression check.
- Iteration safety: snapshotting before mutation is a good reflex even when the spec doesn't strictly require it.

---

### Conclusion

**Approve** — the fix is correct, minimal, and well-scoped. The one substantive finding (misleading `Array.from` comment) is non-blocking; suggest a follow-up commit to either remove the wrapper or correct the rationale, but don't hold the merge for it.

PR #59, opened 2026-04-16
Chart tool: XBRL per company with concept search

nvasilevv:
## Summary

Closes #54.

- **Scope XBRL chart data by company** — the populate endpoint now requires a `ticker` for `xbrl_facts` and uses `query_xbrl_facts_by_ticker` (which joins through Filing → Issuer) instead of querying all facts globally. Previously the XBRL source returned meaningless cross-company data.
- **Add concept discovery endpoint** — `GET /api/charts/xbrl-concepts?ticker=X&keyword=Y` returns available XBRL concepts for a company with fact counts, leveraging the existing `list_xbrl_concepts_for_ticker` repository function and its semantic alias support.
- **Redesigned XBRL data source UI** — when selecting "XBRL Facts" as a data source, users now enter a ticker, search for concepts by keyword, pick from a scrollable results list, and populate the chart — matching the existing UX patterns for EOD prices and analyst estimates.

## Changes

| File | What changed |
|---|---|
| `backend/app/api/charts.py` | New `GET /xbrl-concepts` endpoint; `xbrl_facts` populate now requires ticker and delegates to `query_xbrl_facts_by_ticker` |
| `frontend/src/lib/api.ts` | New `getXbrlConcepts()` API helper |
| `frontend/src/app/[locale]/charts/page.tsx` | Ticker input, concept keyword search with results list, selected concept chip, wired populate call |

## Test plan

- [ ] Select "XBRL Facts" data source on a chart — verify ticker input and concept search appear
- [ ] Enter a known ticker (e.g. FIB) and search concepts — verify results list populates with concept names and counts
- [ ] Click a concept to select it, then populate — verify chart renders with company-scoped XBRL data
- [ ] Verify EOD prices and analyst estimates data sources still work unchanged
- [ ] Verify concept search with keyword filtering (e.g. "revenue", "profit") narrows results

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #63, opened 2026-04-17
Chart data sources: company/concept autocomplete + unsaved-chart flow

nvasilevv:
## Summary
Follow-up to #59. Makes the chart data-source picker usable before the chart is saved, and replaces the static ticker/concept inputs with live autocompletes.

- **Company autocomplete** — shared across EOD Prices, XBRL Facts, and Analyst Estimates. Debounced `searchSecurities` (250 ms) returns ticker + company + exchange, with a `tracked` badge. Selecting a tracked company fills `issuer_id` too, so Analyst Estimates no longer requires pasting a UUID.
- **Concept autocomplete** — XBRL concept suggestions now appear as you type (debounced 300 ms); the separate Search button is gone.
- **Unsaved-chart flow** — the "save the chart first" placeholder is removed. If no chart exists when the user hits Populate, it's auto-created first; the button label switches to "Create & Populate".
- Populate is disabled until the required fields (company, concept for XBRL, issuer_id for estimates) are filled.
- Cleans up a duplicate `/xbrl-concepts` route left behind in #59.

## Test plan
- [ ] Open `/charts`, click +, switch data mode to **Source**, pick **XBRL Facts** — typing a ticker/company shows suggestions; picking one fills the field.
- [ ] Concept field: typing "revenue" shows matching XBRL concepts live; picking one selects it.
- [ ] Click **Create & Populate** on an unsaved chart — chart is created and populated in one step.
- [ ] Repeat for **EOD Prices** (ticker only) and **Analyst Estimates** (tracked issuer only; Populate disabled until a tracked company is picked).
- [ ] Editing an existing chart still works (button reads "Populate").


nvasilevv:
Nothing blocking here, but could you take a pass at the non-blocking items before merge?

**Worth fixing:**
- **Unhandled rejection in the company-autocomplete effect.** Unlike the concept effect, the `searchSecurities` call has no `try/catch` — only `finally`. A network error surfaces as an unhandled rejection. Mirror the concept effect (set `dataError` or swallow).
- **Stale optimistic `selected` on auto-create failure.** In `populate`, `setSelected(created)` runs before `populateChartData`. If populate then fails, the user is silently in "edit existing" mode with an empty chart persisted server-side. Either only `setSelected` after populate succeeds, or delete the freshly-created chart on failure.

**Smaller improvements:**
- Clicking a concept calls `setXbrlKeyword(concept.concept)`, which re-triggers the 300 ms debounced effect and refetches the same list. Harmless but wasteful — a ref flag to suppress the next fetch would avoid it.
- `dsLimit`: `Number(e.target.value) || 100` silently resets on empty input and accepts negative/huge values. A `min`/`max` on the input would help.

**Nits (feel free to ignore):**
- The 150 ms `onBlur` timeout on the company + concept inputs is redundant now that the option buttons use `onMouseDown={e => e.preventDefault()}`. One mechanism is enough; the timeout adds a perceptible lag when tabbing away.
- Composite key `\${sec.exchange}:\${sec.ticker}:\${sec.issuer_id ?? "ls"}` can still collide for two untracked listings with the same exchange+ticker. Index as tiebreaker is fine.
- "No concepts found" can flash between keystroke and debounce firing. Gating on a \`timeoutPending\` flag would smooth it.

Route dedup and the debounced-effect patterns both look great. 👍

nvasilevv:
Thanks — all seven items from the previous round look addressed in 00fac1a4:

- ✅ `searchSecurities` effect now has `try/catch` that clears results and sets `dataError`.
- ✅ Auto-create rollback: `createdHere` flag + `deleteChart(target.id)` on populate failure, and `setSelected` only happens on success via `selectChart(result)`.
- ✅ `skipConceptFetchRef` suppresses the redundant refetch after selecting a concept.
- ✅ `dsLimit` gets `min={1}` / `max={5000}`, rejects non-finite/non-positive, clamps and floors.
- ✅ Both 150 ms `onBlur` timeouts removed (the `onMouseDown` preventDefault is doing the work).
- ✅ Untracked-listing key tiebreaker now `ls-${idx}`.
- ✅ `conceptDebouncePending` gate stops the "No concepts found" flash between keystroke and debounce.

LGTM. 🚢

Issue #65, opened 2026-04-17
Auth: 5 frontend fetch sites bypass apiFetch refresh-on-401


nvasilevv:
## Summary

Five authenticated frontend fetch sites call `fetch()` directly and build the `Authorization` header manually, bypassing the 401 → refresh → retry logic that `apiFetch` in `frontend/src/lib/api.ts` (lines 30-59) provides for every other endpoint. When the 15-minute access token expires, these requests fail silently or surface misleading errors instead of transparently refreshing and retrying.

This is a follow-up to #26 (chat "token expired" mid-conversation), which has the same root cause but requires a different fix because the chat uses a custom SSE interceptor fetch. These 5 sites, by contrast, are a mechanical refactor.

## Affected sites

| File | Endpoint | Current symptom when access token is expired |
|---|---|---|
| `frontend/src/hooks/useLiveFilings.ts:30` | `GET /api/filings` (polled) | Feed silently stops updating; user sees stale data |
| `frontend/src/components/pipeline/pipeline-dashboard.tsx:110` | `GET /api/ingestion/pipeline` (polled) | Dashboard shows \"HTTP 401\" error text |
| `frontend/src/components/pipeline/pipeline-dashboard.tsx:504` | `POST /api/ingestion/retry-dead-letter/{id}` | Retry button silently no-ops |
| `frontend/src/components/graph/knowledge-graph.tsx:54` | `GET /api/notes/graph/full` | Graph renders empty |
| `frontend/src/components/generative/chart-renderer.tsx:50` | `GET /api/charts/{id}` | Chart shows \"Chart not found\" |

All five follow the same pattern:

```ts
const token = getToken();
const headers: Record<string, string> = {};
if (token) headers.Authorization = `Bearer ${token}`;
const res = await fetch(`${API_BASE_URL}/api/...`, { headers });
```

## Fix

Replace each call site with `apiFetch` from `frontend/src/lib/api.ts`. `apiFetch` already:

- Injects the `Authorization: Bearer <token>` header from `getToken()`
- On 401, calls `POST /api/auth/refresh` (deduped via a module-level `refreshPromise`)
- Retries the original request with the new token
- Logs the user out and redirects to sign-in if refresh fails

The diff at each site is roughly:
- Remove the `getToken() + headers` preamble
- Swap `fetch(url, init)` → `apiFetch(url, init)` (may need to export `apiFetch` from `lib/api.ts` if not already — currently it is not exported).

## Out of scope / already handled

[...]

PR #67, opened 2026-04-17
Fix chat "token expired" mid-conversation (#26)

### LINKED ISSUE #26: Chat returns token expired in the middle of conversation
Token expired is returned in the chat from time to time while in the middle of the chat. Once message resent, things work fine.

nvasilevv:
Closes #26.

## Problem

Users intermittently saw "Token expired" surfaced in the chat UI mid-conversation. Resending the message worked, which made it look flaky rather than a real auth issue.

## Root cause

The chat streaming transport (`DefaultChatTransport` in `frontend/src/components/feed/ask-panel.tsx`) uses a custom fetch — `createInterceptingFetch` — so it can pull custom SSE events (chat metadata, citation maps, step starts) out of the stream before the AI SDK parser sees them. That custom fetch calls `globalThis.fetch` directly and therefore bypassed the `401 → refresh → retry` logic that `apiFetch` in `frontend/src/lib/api.ts` provides for every other authed endpoint.

When the 15-minute access token expired between chat turns, the next send to `/api/ask/stream` hit the backend with a stale token, got `401 {"detail": "Token expired"}`, and the error bubbled up to the UI. "Resend fixes it" only worked when something else (a REST call via `apiFetch`) happened to have refreshed the token in between.

## Fix

- Extract the existing deduped refresh logic in `api.ts` into a shared `refreshAccessToken()` helper and export it. `apiFetch` now calls the helper instead of inlining the dedup — no behavior change for REST callers.
- In `createInterceptingFetch`, if the initial response is 401, call `refreshAccessToken()`, rewrite the `Authorization` header on `init`, and retry once before entering the SSE reader loop. If refresh fails, fall through to the existing error path (which triggers logout on the next `apiFetch` call).

The `refreshPromise` module-level singleton in `api.ts` means concurrent 401s (e.g. chat fetch + a REST poll) share a single refresh round-trip, so the new caller doesn't add pressure on the refresh endpoint.

## Scope

Only the chat streaming fetch is fixed here. The same pattern (direct `fetch` bypassing `apiFetch`) exists in 5 other spots in the frontend — filed as #65 for a separate, mechanical refactor PR since those can just swap `fetch` → `apiFetch` and don't need the shared helper.

## Test plan

- [ ] In `backend/app/core/config.py`, temporarily set `access_token_expire_minutes = 1`, restart backend.
- [ ] Log in, send a chat message (works as today).
- [ ] Wait >1 minute idle, send another chat message → expected: streams normally, no "Token expired" shown.
- [ ] Clear the `refresh_token` cookie in devtools, send a chat message → expected: clean logout + redirect to sign-in (unchanged behavior).
- [ ] Revert the `access_token_expire_minutes` change before merging or restarting staging.
- [ ] `tsc --noEmit` on `frontend/` passes locally.


nvasilevv:
# PR Review

## Summary
Extracts the existing deduped refresh logic from `apiFetch` into an exported `refreshAccessToken()` helper and wires it into `createInterceptingFetch` in `ask-panel.tsx` so the chat SSE stream gets the same `401 → refresh → retry` treatment as every other authed request. The change is small, targeted, and faithfully mirrors the behavior it's borrowing from.

## Review Findings

### 🟡 Improvement — please confirm before merging

Both of these need an explicit check from the author — not optional polish:

1. **Retry loses headers when `input` is a `Request` object.** In the retry branch (`ask-panel.tsx:47-51`), `new Headers(init?.headers)` only picks up headers that were passed through `init`. If the caller (AI SDK's `DefaultChatTransport`) ever hands in a `Request` object with headers baked in, those headers are dropped on the retry and only `Authorization` is sent — `Content-Type`, `Accept`, any custom headers, gone. Today `DefaultChatTransport` happens to use `init.headers`, so this is latent rather than broken, but that's an implementation detail of the SDK version we're pinned to. **Please verify against the current `ai` package version that `input` is always a string/URL here, or defensively merge headers from `input instanceof Request ? input.headers : undefined` as well.** Leaving this unverified means a future SDK bump can silently break chat auth.

2. **No logout/redirect on unrecoverable 401 in the streaming path.** When `refreshAccessToken()` returns `null`, `apiFetch` tears down the session (`api.ts:49-54`: logout + redirect). The intercepting fetch just lets the original 401 response fall through into the SSE reader. The PR body acknowledges this (\"triggers logout on the next apiFetch call\"), but that leaves the user in a broken chat state — stream errors out, no redirect — until some unrelated REST call happens to fire. **Please either mirror the logout branch here (three lines) or explicitly justify why the deferred logout is acceptable UX.** Right now the behavior is \"it kind of self-heals eventually,\" which is exactly the class of flakiness #26 was filed to eliminate.

### 🔵 Nitpick
- The exported helper's one-line comment (`api.ts:30`) duplicates the old inline comment. Since `refreshAccessToken` is now the public surface, add a note that `null` return = refresh rejected, caller should treat session as dead.
- If a third caller shows up (per #65), consider extracting a `withRefreshedAuth(init, token)` helper so the three-line header-rewrite dance doesn't get copy-pasted again. Not worth blocking on.

### Security & Correctness
- No injection, secret, or authz concerns. Refresh endpoint unchanged, `credentials: \"include\"` preserved, token storage path unchanged.
- Dedup singleton still guarantees at most one in-flight refresh across REST + streaming callers.
- Retry only fires on `response.status === 401`, so no risk of infinite loops or retry storms.

## Positive Feedback
- Extracting `refreshAccessToken` without changing `apiFetch`'s observable behavior is exactly the right shape — minimal blast radius, reuses the battle-tested dedup, and sets up #65 cleanly.
- Excellent PR body: clear root cause, explicit scope boundary (#65), reproducible test plan with the `access_token_expire_minutes = 1` trick.
- Keeping the fix surgical rather than bulk-migrating the other 5 `fetch` sites is the right judgment call.

## Conclusion
**Approve pending confirmation on the two 🟡 items above.** The fix itself is correct and well-scoped; I just want the author to explicitly address (or acknowledge and defer) those two before this lands.

nvasilevv:
Both 🟡 items addressed cleanly in 651f2bd — header merge precedence is correct and the logout branch mirrors apiFetch. Approve, ready to merge.

PR #69, opened 2026-04-17
Fix chat stream crash on second message (DetachedInstanceError)

### LINKED ISSUE #66: Chat agent drops interactions after first message response
Or even the chat, the agent is basically not responding to the second question in a chat and basically no other question afterwards.

nvasilevv:
## Summary

Fixes #66 — the chat agent stopped responding after the first message.

The `/ask/stream` producer was reading `chat_row.title` after the `get_session_context()` block had closed. SQLAlchemy expires attributes on commit and detaches the instance when the session closes, so that access raised `DetachedInstanceError` and killed the stream silently — the HTTP response returned 200 but no SSE events reached the client, making the agent appear unresponsive to the second (and every subsequent) message in a chat.

The first message worked because in that branch a fresh `Chat` instance is created in Python and `.title` was assigned before any commit, so the attribute was still in the instance `__dict__`. The `else` branch (existing chat_id, i.e. every follow-up turn) fetched the row, committed `updated_at`, expired the attributes, exited the session, and then tried to read `.title` — boom.

Production logs on `ceewire-api` confirmed the exact trace at `routes.py:2919` on every repro.

## Changes

- Capture `chat_title` as a plain `str` inside the session context.
- Use that local when emitting the `chat-metadata` SSE event.
- Move the "`chat_row` unexpectedly None" warning into the session block where the row was actually fetched.

No schema or contract change — the SSE payload is identical.

## Test plan

- [ ] Deploy the branch and run a 3+ turn conversation; confirm each turn streams a response.
- [ ] Check `docker logs ceewire-api` — no more `DetachedInstanceError` tracebacks after turn 1.
- [ ] Verify the `chat-metadata` event still carries a `title` on both new and existing chats (new chat: derived from first user message; existing chat: persisted title).


nvasilevv:
## Summary

Fixes a silent stream crash where every chat message after the first triggered a `DetachedInstanceError` because `chat_row.title` was read after its SQLAlchemy session had closed. The fix captures the title as a plain `str` inside the `get_session_context()` block and uses that local when emitting `chat-metadata`. Tight, well-scoped change with a clear root-cause writeup in the PR body.

## Review Findings

### 🔴 Critical/Security
None.

### 🟡 Improvement
- **`backend/app/api/routes.py:2897` — silent fallthrough when an unknown `chat_id` is passed.** When the `else` branch fetches and `chat_row` is `None`, the new code logs a warning but still proceeds: `_chat_id` keeps the client-supplied UUID, the user message gets inserted with a `chat_id` that has no parent `Chat` row (likely an FK violation depending on schema), and `chat-metadata` is emitted with `title=""`. The pre-fix code had the same gap, so this isn't a regression — but while you're here, consider returning an SSE error event and aborting, instead of writing orphan rows. Also worth checking whether the row is missing because it belongs to a different `user_email` (auth bypass smell) vs. genuinely absent.
- **Comment is slightly load-bearing for future readers.** The new comment is good, but the *reason* the new-chat branch never triggered the bug (attribute lives in `__dict__` because it was assigned in Python before commit, so expiration is moot until first access post-detach) is the kind of subtlety someone will re-break in 6 months. Worth one extra sentence, or a regression test that exercises the second-message path against a real session.

### 🔵 Nitpick
- `chat_title: str = \"\"` — the annotation is fine but redundant given the literal initializer; the surrounding code doesn't annotate similar locals. Minor consistency thing.
- The `else` block now mixes \"refresh updated_at\" with \"capture title for SSE.\" Reads fine, but a one-line `# capture before session closes` next to the `chat_title = ...` line would mirror the upper comment and make the symmetry obvious.

## Positive Feedback
- Excellent PR description: identifies the exact line, explains why the first-message path masked the bug, and cites the production trace. This is the gold standard for a one-line fix.
- Moving the `chat_row is None` warning *into* the session block is the right call — it now fires where the row was actually missing, instead of after a successful refresh path where `chat_row` could only be `None` if the prior block was buggy.
- No contract change to the SSE payload — frontend is unaffected.

## Conclusion
**Approve.** The fix is correct, minimal, and well-justified. The orphan-write concern in the unknown-`chat_id` path is pre-existing and worth a follow-up issue rather than blocking this PR.

PR #82, opened 2026-04-20
Fix dashboard layout ratio: charts 1/3, filings 2/3

### LINKED ISSUE #70: Landing page in app -fix the ratio for charts on first line and list of filings on second
Current ratio looks more two-thirds for the charts and one-third for the filings, and it should be the other way around.

nvasilevv:
Closes #70

## Summary
- On large screens (`lg`), `MarketPulse` (charts) had `shrink-0` with no height cap, causing it to consume ~2/3 of the vertical space while `LiveFeed` (filings) only got ~1/3
- Added `lg:basis-1/3 lg:overflow-hidden` to the MarketPulse wrapper so charts are constrained to 1/3 of the container height
- `LiveFeed` already has `flex-1`, so it automatically claims the remaining 2/3
- Mobile layout is unaffected (scrollable flex column, no `overflow-hidden`)

## Test plan
- [ ] Open the dashboard on a large screen (≥1024px) and verify charts take ~1/3 height, filings ~2/3
- [ ] Verify mobile layout is unaffected

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


nvasilevv:
## Summary

PR #82 claims to fix dashboard layout (issue #70) but the diff contains none of that work — it actually fixes issue #28 (new chats not appearing dynamically in the history list). The real change introduces an optimistic chat-list update mechanism via a renamed `chatListChanged` CustomEvent with `{id, title}` payload, extends the SSE `chat-metadata` event with the chat title, and adds a debounced background refetch plus a failure-distinguishing `getChats(): … | null`.

## Review Findings

### 🔴 Critical

- **PR title and description are wrong.** Title says "Fix dashboard layout ratio: charts 1/3, filings 2/3"; body says "Closes #70"; test plan describes the dashboard. None of that code is in the diff. The real change is the chat-list optimistic-update fix for #28 (confirmed by commit history: `Fix new chats not appearing dynamically in the chat history list (#28)`). Merging this as-is would close the wrong issue, leave #28 open, and contaminate the git log. **Rewrite the PR title/body (and the closing-issue reference) to describe the actual change before merging.**

### 🟡 Improvement

- **`setChats(result)` can clobber an optimistic entry mid-flight.** `chat-history.tsx:148` blindly replaces state with the server response. If `fetchChats()` was in flight when a `chatListChanged` event fired and inserted an optimistic row, the resolving fetch (which may predate the new chat commit) overwrites the optimistic row and drops it. The debounce narrows the window but doesn't close it — under load the server read can still race the create commit. Consider merging: keep any local entry whose `id` isn't in `result` (i.e., upsert server rows onto the current state instead of replacing).
- **Initial-load retry has no terminal error UX.** `chat-history.tsx:160-174` does one blind 1.5s retry then flips `loading=false` regardless of success. A genuinely failed load (expired token, 5xx) leaves an empty panel indistinguishable from "no chats yet" (`chat-history.tsx:283`). At minimum, track the failure and surface a retry affordance, or log.
- **Backend `chat_row is None` path emits an empty-title event rather than erroring.** `routes.py:2914-2921` logs a warning but then proceeds to stream `title=""`. If the `_chat_id` truly isn't in the DB, the rest of the stream will likely also be broken (the user message was saved with a dangling FK at line 2903-2909 — which itself may have 500'd, depending on schema). The warning is the right instinct; the handling isn't. Either bail out cleanly with an error SSE event, or remove the conditional (if it's truly impossible, `assert` and let the exception surface).

### 🔵 Nitpick

- **Long explanatory comment in `ask-panel.tsx:509-512`** (about "second of two intentional dispatches") is load-bearing for future readers given the dual-dispatch design — keep, but consider moving it to a single place near the handler in `chat-history.tsx` so it isn't duplicated in spirit across the three dispatch sites.
- **`_meta: dict[str, Any]` annotation** at `routes.py:2916` is unnecessary — `_sse` accepts it fine without the hint. Minor.
- **`UNTITLED_CHAT` constant** is nice; consider co-locating with the `ChatListItem` type if other components also render an untitled fallback.

## Positive Feedback

- The failure-distinguishing `getChats(): ChatListItem[] | null` change is exactly the right call — returning `[]` on failure was the latent bug that would have made any naive "just refetch" fix still overwrite optimistic state. Single-caller, so the signature change is safe.
- Comments on the debounce rationale (`chat-history.tsx:140-141`) and the Safari `start()`-vs-`pull()` workaround elsewhere in this file are earning their keep — they explain non-obvious constraints that would otherwise be re-debugged.
- Upsert-by-id in the event handler (`chat-history.tsx:182-197`) cleanly handles both the pre-navigation dispatch and the later SSE metadata dispatch without dedup logic on the caller side.

## Conclusion

**Request Changes.** The code change itself is close to mergeable (small data-race concern in the refetch merge and a backend error-handling nit), but the PR metadata is inaccurate in a way that will corrupt issue tracking and history if merged. Fix the title/body/closing-issue reference, then address the `setChats` race and the `chat_row is None` handling.

nvasilevv:
Opened against wrong branch (pr-50 instead of issue-70-fix-dashboard-ratio). Replacing with correct PR.

PR #83, opened 2026-04-20
Fix source file download stalling in mobile PWA

### LINKED ISSUE #71: Download of source files doesnt work
so there must be something off with the download of the source document and here I'm talking about the files that are available when you click on citations from the agent and there is a download button to download the actual source files. I tried that in my latest chat about shelly group with the agent and it just stalls. I mean it opens up a new window and nothing happens. It's on mobile. Haven't tested it on my laptop yet but just wanted to make sure that nothing is broken there

nvasilevv:
## Summary
- Closes #71 — download button on citations opens a blank window on iOS PWA and never actually downloads.
- Root cause: the `<a target="_blank">` points to a cross-origin R2 presigned URL that returns no `Content-Disposition`; Safari tries to render binary files inline and stalls silently, and the cross-origin `download` attribute hint is ignored by the browser.
- Fix: make R2 return `Content-Disposition: attachment` and an explicit `Content-Type` via the `ResponseContent*` overrides in `generate_presigned_url`. Every browser (desktop Safari, mobile Safari, standalone PWA) now treats the response as a download.

## Changes
- `backend/app/services/storage.py` — `generate_presigned_url` gains optional `download_filename` and `content_type` kwargs. Encodes the filename as RFC 5987 (`filename*=UTF-8''…`) with a quoted ASCII fallback and strips any path components to prevent header injection.
- `backend/app/api/routes.py` — all four call sites now pass a filename + content type: the citation endpoint (`/documents/content`), the reports list, the report download redirect, and the `generate_report` agent tool.
- `backend/tests/test_storage_presigned_url.py` — 4 unit tests covering: no-filename baseline, ASCII + UTF-8 disposition forms, non-ASCII percent-encoding, and path-traversal stripping.
- `frontend/src/components/feed/document-viewer.tsx` — harmless `download={filename}` hint on the two citation download anchors (browsers ignore it cross-origin, but keeps intent clear).

## Test plan
- [x] `pytest backend/tests/test_storage_presigned_url.py` — 4/4 pass
- [ ] Deploy, then verify on iOS PWA: click a citation → Download button → file actually downloads (tests cannot simulate Safari View Controller)
- [ ] Sanity-check on desktop Safari (Mac) that downloads still work (no regression)
- [ ] Sanity-check the "Download generated report" flow from the reports list (also uses `generate_presigned_url`)


nvasilevv:
Drive-by review — all non-blocking, fine to defer or fold into a follow-up:

- **`ascii_fallback` doesn't strip control characters** (`backend/app/services/storage.py`). Only `"` is removed; CR/LF/other control bytes would pass through into the `Content-Disposition` value. boto3 URL-encodes the query param and filenames are DB-sourced, so header injection isn't a live risk — just defense-in-depth. A `"".join(c for c in ascii_fallback if c.isprintable())` closes it.
- **Duplicated basename expression** (`backend/app/api/routes.py:858,878`). `r.storage_key.rsplit("/", 1)[-1]` appears twice for the reports-list and reports-download endpoints; `Path(r.storage_key).name` or a small helper would keep them in lockstep.
- **Missing test for `content_type` without `download_filename`** (`backend/tests/test_storage_presigned_url.py`). The two kwargs are independent but only the filename-present path is covered.
- **Frontend `download={filename}` is a no-op on cross-origin anchors** (`frontend/src/components/feed/document-viewer.tsx`). PR body already acknowledges this — flagging only because future readers may assume it's load-bearing.

Approving as-is.

PR #84, opened 2026-04-20
Fix dashboard layout ratio: charts 1/3, filings 2/3

### LINKED ISSUE #70: Landing page in app -fix the ratio for charts on first line and list of filings on second

nvasilevv:
Closes #70

## Summary
- On large screens (`lg`), `MarketPulse` (charts) had `shrink-0` with no height cap, causing it to consume ~2/3 of the vertical space while `LiveFeed` (filings) only got ~1/3
- Added `lg:basis-1/3 lg:overflow-hidden` to the MarketPulse wrapper so charts are constrained to 1/3 of the container height
- `LiveFeed` already has `flex-1`, so it automatically claims the remaining 2/3
- Mobile layout is unaffected (scrollable flex column, no `overflow-hidden`)

## Test plan
- [ ] Open the dashboard on a large screen (≥1024px) and verify charts take ~1/3 height, filings ~2/3
- [ ] Verify mobile layout is unaffected

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


nvasilevv:
## Summary

PR #84 is a re-do of what PR #82's description *claimed* to do — a single-line class change in `dashboard/page.tsx` that caps `MarketPulse` (charts) at 1/3 height on `lg+` screens so `LiveFeed` (which already has `flex-1`) claims the remaining 2/3. Title, body, branch name (`issue-70-fix-dashboard-ratio`) and diff are finally consistent with #70.

## Review Findings

### 🟡 Improvement

- **Short viewports will clip chart content.** `MarketPulse` (`market-pulse.tsx:62-68`) renders skeletons at `lg:h-[240px]` and `ResponsiveContainer height={160}` + header + legend + paddings → each card is ≈220–260 px tall. With `lg:overflow-hidden` on the wrapper and `lg:basis-1/3`, a 720p laptop (1/3 of ~680 px usable = ~227 px) will hard-clip the bottom of the Sentiment bars and "Most Active" rows. Two ways to mitigate:
  1. Replace `lg:overflow-hidden` with `lg:overflow-y-auto` so tall content is reachable via scroll inside the panel. Preserves the ratio while staying usable on short screens.
  2. Or make the chart cards internally scroll (`overflow-y-auto` on the 3rd card's inner list) so the ratio is hard but content stays reachable.
- **Fixed 1/3 ignores breakpoints above `lg`.** On ultra-wide 4K (`2xl`+), 2/3 of the screen to filings is fine, but consider whether `xl:basis-1/4` would read better when the chart cards have ample horizontal room. Optional — product call.
- **`shrink-0` + `basis-1/3` is deliberate but worth a comment.** A reader glancing at the three Tailwind utilities (`shrink-0 lg:basis-1/3 lg:overflow-hidden`) might "simplify" by dropping `shrink-0`, which would let the wrapper shrink below 1/3 when `LiveFeed` content pushes. If you feel the why is non-obvious, a one-line comment above the div earns its keep; otherwise fine.

### 🔵 Nitpick

- PR body says "charts are constrained to 1/3 of the container height" — slightly imprecise. `basis-1/3` + `shrink-0` fixes the basis at 33.333% of the flex parent's main-axis size, but the filter bar (`page.tsx:34-66`) and padding share the same axis, so in practice `LiveFeed` gets ~(2/3 − filter_bar_height). Close enough, but worth a note if the exact ratio matters.
- Test plan is still unchecked boxes — presumably you'll tick them after local verification; not blocking.

## Positive Feedback

- Minimal, surgical diff — one line changed, zero collateral. Exactly the right scope for a UI ratio fix.
- Branch name, title, body, and closing-issue reference are all aligned with the diff. This is what #82 should have looked like.
- Correct choice to constrain `MarketPulse` (which has `shrink-0` and an uncapped intrinsic height) rather than trying to boost `LiveFeed`'s `flex-grow` — the root cause was the unbounded chart panel, not an under-claiming feed.

## Conclusion

**Approve** (with the short-viewport clipping as a follow-up worth considering — not a blocker if the target audience is on ≥1080p displays). The PR does exactly one thing, does it correctly, and its metadata matches its diff.

nvasilevv:
Follow-up review on `fdb50fd`:

New commit swapped `lg:overflow-hidden` → `lg:overflow-y-auto` — exactly the mitigation suggested (option 1). Short-viewport clipping is resolved: tall chart content is now reachable via scroll inside the 1/3 panel.

**What's left:**

- 🔵 PR body still says *"Added `lg:basis-1/3 lg:overflow-hidden`"* — stale after the follow-up commit. Minor doc drift, not blocking.
- 🔵 Test plan checkboxes still unticked.
- 🟡 The `xl:basis-1/4` ultra-wide consideration — optional product call, skip unless relevant.
- 🟡 One-line comment explaining the `shrink-0` + `basis-1/3` combo — still optional.

Nothing blocking. The substantive concern (clipping) is addressed; remaining items are cosmetic. **Approve.**

PR #85, opened 2026-04-20
Add resend button under user messages in chat

### LINKED ISSUE #72: Chat messages - resend
It would be nice to add a recent button underneath each message from the user when chatting with the agent. This will be easier to read and write messages for whatever reason

nvasilevv:
## Summary
- Adds a small resend button in the hover-action row under each user message in the agent chat (next to the existing copy button).
- Clicking it re-submits the same text via `sendMessage({ text })` — the same call `onSubmit` uses.
- Button is disabled while the agent is streaming to prevent stacking submissions.

Closes #72

## Test plan
- [ ] `cd frontend && npm run dev`, open the agent chat
- [ ] Send a message, hover the user bubble, confirm a resend icon appears next to the copy icon
- [ ] Click resend — same text should be submitted as a new user message and the agent should respond
- [ ] While the agent is still streaming, confirm the resend button is visibly disabled and does not fire


nvasilevv:
# PR Review

## Summary
Small UI change: adds a `ResendButton` next to `CopyButton` under each user message in `AskPanel`, wired to `sendMessage({ text })` and disabled while `isStreaming`. Scope is contained to `frontend/src/components/feed/ask-panel.tsx` (+20/-0).

## Review Findings

### 🟡 Improvement

- **Resend path diverges from `onSubmit` setup (ask-panel.tsx:822-832 vs 978-980).** `onSubmit` performs four pieces of bookkeeping before calling `sendMessage`: `setChatError(null)`, `setSentHistory(prev => [...prev, text])`, `setHistoryIndex(-1)`, `setStepCount(0)`. The resend handler skips all of them. Most load-bearing is `setStepCount(0)` — the UI at line 925 conditions on `isStreaming && stepCount > 0`, so a stale step count from the prior turn can leak into the new stream's first frames. `setChatError(null)` matters if the user is resending because the previous message errored (a likely real use case). Consider calling `onSubmit` after setting the input, or extracting a `submitText(text: string)` helper that both code paths call.

- **`sentHistory` not updated on resend (ask-panel.tsx:826).** Arrow-up recall won't see resent messages. Debatable (the user didn't type it), but inconsistent — and cheap to fix if you extract the helper above.

### 🔵 Nitpick

- **`ResendButton` API is slightly over-parameterized (ask-panel.tsx:458).** The only call site passes `(t) => sendMessage({ text: t })`, and `text` is already a prop. You could drop `onResend` and accept a plain `onClick`, or inline the button entirely given it's used once. Also creates a new function on every render of the user-message row.
- **Missing `aria-label` (ask-panel.tsx:462).** `title` is a tooltip, not an accessible name. `CopyButton` has the same gap, so you may want to address both together — but if scoping strictly, this is new code.
- **Long className literal (ask-panel.tsx:463).** Readable but dense; a wrapped multi-line string would match patterns elsewhere in the file.

### 🔴 Critical/Security
None. `text` flows through existing React state; no new injection or data-exposure surface.

## Positive Feedback
- Reuses the already-imported `RotateCcw` icon — no new imports (ask-panel.tsx:8).
- Styling and placement mirror `CopyButton` exactly, keeping the hover-action row visually consistent.
- Correctly gates on `isStreaming`, which already covers both `"streaming"` and `"submitted"` (ask-panel.tsx:555), so double-submits during the early submit window are handled.

## Conclusion
**Comment.** The change is small and low-risk, but I'd like to see the resend path share `onSubmit`'s reset logic (at minimum `setStepCount(0)` and `setChatError(null)`) before merging — extracting a shared `submitText` helper is the cleanest fix.

nvasilevv:
## Follow-up review on 813e96d

Both 🟡 concerns and two of three 🔵 nitpicks are addressed.

**Addressed:**
- 🟡 **Shared reset logic**: extracted `submitText(text)` helper (ask-panel.tsx:839-845) that does all four resets + `sendMessage`. Both `onSubmit` and the resend button call it, so `setStepCount(0)` / `setChatError(null)` / history bookkeeping now apply on resend.
- 🟡 **sentHistory on resend**: implicitly fixed — resend now appends to history via the shared helper. Arrow-up recall will see resent messages.
- 🔵 **`ResendButton` API**: simplified to `{ onClick, disabled }` (ask-panel.tsx:458). Cleaner, no pass-through `text` prop.
- 🔵 **`aria-label`**: added on the resend button (ask-panel.tsx:464).

**Not addressed (intentional/acceptable):**
- 🔵 Long className literal — unchanged, matches existing `CopyButton` style.
- The inline `() => submitText(text)` still allocates per render, same as the old form. Negligible.

**Note on the refactor:** the split between `onSubmit` (input-specific: clears input, resets textarea height) and `submitText` (send-path bookkeeping) is clean — `onSubmit` correctly stays the caller for input-driven submits, and resend goes straight to `submitText` without needing input state. Good factoring.

Flipping to **Approve**.

PR #86, opened 2026-04-20
Fit Market Pulse cards into 1/3 panel (follow-up to #84)

nvasilevv:
## Summary
- Follow-up to #84. After that PR capped the chart panel at 1/3 of viewport height, the three Market Pulse cards overflowed, and a follow-up commit worked around it by making the whole panel scroll (`lg:overflow-y-auto`). This replaces that workaround: cards now shrink to fit the 1/3 slot.
- **`dashboard/page.tsx`**: `lg:overflow-y-auto` → `lg:overflow-hidden`, added `lg:min-h-0`.
- **`market-pulse.tsx`**:
  - Grid + all 4 cards claim full panel height (`lg:h-full lg:min-h-0`); grid drops `mb-3` on `lg` so the margin doesn't eat into the allocation.
  - Card 3 (Most Active) gets `flex flex-col` (was missing) and row density drops from `py-1.5` to `py-0.5` so all 10 rows fit. Inner list keeps `overflow-y-auto` as a fallback for very short viewports.
  - Card 1 (Filing Activity): chart replaced fixed `height={160}` with `height="100%"` inside a `flex-1 min-h-0` wrapper so it fills remaining vertical space.
  - Loading skeleton: `h-[180px] lg:h-[240px]` → `h-[180px] lg:h-full` to match.

## Test plan
- [ ] Dashboard at 1280×720 and 1920×1080: charts panel occupies ~33% height, all three cards fit, no scrollbar.
- [ ] Filing Activity chart renders with visible axes and ticks at the new shrunken height.
- [ ] Sentiment bars centered and all three visible.
- [ ] Most Active shows all 10 rows without scroll at normal heights.
- [ ] Short viewport (1280×650): cards still fit; Most Active inner scroll may engage — acceptable.
- [ ] Mobile (375×812): unchanged — cards stack, page scrolls normally.
- [ ] Light and dark themes both render correctly.


nvasilevv:
## Summary
Small, focused layout fix that replaces a prior `overflow-y-auto` workaround (added after #84 capped the Market Pulse panel at 1/3 viewport height) with proper flex/grid `h-full` + `min-h-0` sizing so the three cards shrink to fit their slot. Filing Activity chart switches from a fixed 160px to percent-based `ResponsiveContainer`, Most Active drops row density from `py-1.5` to `py-0.5` to fit 10 rows, and the loading skeleton is updated to match.

## Review Findings

| Severity | Finding |
|---|---|
| 🔴 | None. No security, auth, or data-handling changes — purely CSS. |
| 🟡 | `market-pulse.tsx` inner list uses `overflow-y-auto` as a fallback for short viewports. Good, but `py-0.5` (4px total) is quite tight. At narrower widths or with denser fonts the ticker/company/count row may feel cramped. Worth verifying visually at 1280×650 the list actually fits without engaging the scroll — if it doesn't, the density drop was cosmetic only. Consider `py-1` as a middle ground if readability degrades. |
| 🟡 | `market-pulse.tsx` — after wrapping `<ResponsiveContainer>` in a new `<div class=\"flex-1 min-h-0\">`, the inner chart children (`<CartesianGrid>`, `<XAxis>`, etc.) were not re-indented. They're now one additional nesting level deeper but sit at the old column, so the JSX tree no longer matches its indentation. Reindent for readability and to keep diffs clean on future edits. |
| 🔵 | Duplicate class list `mb-3 grid grid-cols-1 gap-3 lg:mb-0 lg:grid-cols-3 lg:h-full` appears on both the skeleton and real grid. Minor DRY opportunity — extract to a const if you touch this again, but not worth a dedicated change. |
| 🔵 | `lg:mb-0` is used to zero out the `mb-3` margin at `lg` so it doesn't eat into the 1/3 allocation. Slightly clearer alternative: drop `mb-3` at the panel level and let the parent control spacing. Stylistic. |
| 🔵 | `height=\"100%\"` on `ResponsiveContainer` inside `flex-1 min-h-0` is the correct Recharts pattern. On the very first paint there can be a brief measurement pass where the chart renders at 0 height before the parent settles; if you see a flash, the usual fix is `minHeight={120}` on the wrapper div — not needed to ship. |

## Positive Feedback
- Nice root-cause fix: replacing `overflow-y-auto` on the panel with proper `h-full` + `min-h-0` propagation down the grid/flex chain is exactly the right Tailwind idiom for \"shrink to container, let children scroll if needed.\"
- Loading skeleton updated in lockstep with the real layout (`h-[180px] lg:h-full lg:min-h-0`) — easy to forget, and skipping it would cause layout pop on first load.
- Clear, specific PR description with explicit before/after reasoning per card.
- Inner `overflow-y-auto` on Most Active as a short-viewport fallback is a thoughtful guardrail rather than a primary solution.

## Conclusion
**Approve** — ship it, but please address the nitpicks above before merging: the JSX reindent inside the chart wrapper, the `py-0.5` visual check at 1280×650, and a quick look at the minor DRY/margin notes. Thanks!

nvasilevv:
Thanks for the review! Addressed in a36e603:

- 🟡 **JSX re-indent**: AreaChart children now at the correct nesting level under the new `flex-1 min-h-0` wrapper.
- 🟡 **`py-0.5` density at 1280×650**: I can't run a browser here, so leaving to your visual check — if the 10 rows feel cramped at that viewport, swap `py-0.5` → `py-1` on the row div in `market-pulse.tsx`. At `py-1` 10 rows are right at the ~210px budget, so `overflow-y-auto` will likely engage on shorter viewports.
- 🔵 **DRY grid class / mb-3 / ResponsiveContainer flash**: noted, leaving as-is to keep the diff focused.

PR #87, opened 2026-04-20
Fix notes graph close button hidden behind iOS status bar

### LINKED ISSUE #73: Misalignment of controls for notes graph
There is misalignment of the control when opening the graph for research notes on a mobile device. I'm here testing with my phone. When I open it up, what happens is that the closing button (the little x at the top right corner) is actually hidden behind my clock so I cannot close the window unless I go into landscape mode. I want to fix that

nvasilevv:
## Summary
- Adds `env(safe-area-inset-top)` padding to the Knowledge Graph panel header so the close (X) button isn't obscured by the iOS status bar / clock in portrait mode.
- Reuses the exact pattern already established in `chat-context-panel.tsx` — `max(0.75rem, env(safe-area-inset-top, 0.75rem))` keeps current desktop spacing (`py-3` = 0.75rem) and only grows on devices that report a safe-area inset.

Fixes #73

## Test plan
- [ ] Open Knowledge Graph on an iOS device in portrait (installed PWA or Safari) — close button is visible below the status bar clock and tappable.
- [ ] Rotate to landscape — close button still renders correctly.
- [ ] Desktop browser — header padding is unchanged.


nvasilevv:
## Summary
Adds `env(safe-area-inset-top)` padding to the Knowledge Graph panel header so the close (X) button isn't hidden by the iOS status bar in portrait. The change reuses the pattern already established in `chat-context-panel.tsx:185` and relies on `max()` to preserve existing `py-3` (0.75rem) spacing on desktop. `viewportFit: "cover"` is already set in `app/layout.tsx:18`, so the inset will actually be non-zero on notched devices.

## Review Findings

### 🔴 Critical/Security
None.

### 🟡 Improvement
- **`frontend/src/components/graph/graph-panel.tsx:56-58`** — The chat-context panel gates the safe-area padding with `md:hidden` because its close button only renders on mobile. Here, the graph panel is rendered on desktop too, so the rule is applied unconditionally. It's functionally safe — `max(0.75rem, 0)` collapses to `0.75rem` on desktop — but it means any future tweak to desktop header height has to remember this inline override exists. Consider either a comment noting the intent, or applying the style only when the viewport actually needs it (e.g., a utility class `pt-[max(0.75rem,env(safe-area-inset-top,0.75rem))]` keeps the value co-located with the other Tailwind classes).

### 🔵 Nitpick
- **`graph-panel.tsx:57`** — Mixing inline `style` with Tailwind's `py-3` works (inline wins for `padding-top` only, leaving `padding-bottom` from `py-3`), but it's a slightly subtle interaction. A Tailwind arbitrary value would keep everything in one place and match the rest of the header's styling approach. Minor — the existing `chat-context-panel` uses the same inline pattern, so leaving it matches convention.
- Test plan is unchecked — presumably still pending device verification. Worth confirming on a real notched iOS device before merge, since there's no automated coverage for this.

## Positive Feedback
- Tight, minimal, targeted fix that mirrors the existing project pattern rather than inventing a new one. The `max(baseline, env(..., baseline))` idiom is the right construction — it keeps desktop visually unchanged while guaranteeing at least the status-bar inset on iOS, and degrades safely on browsers that ignore `env()`.
- Good linkage to the underlying issue (#73) with clear reproduction steps.

## Conclusion
**Approve** — once the iOS portrait test in the test plan is actually performed. The change is correct, minimal, and consistent with the codebase's existing safe-area handling.

nvasilevv:
Follow-up: both concerns from the previous review are addressed in `f2918bd`.

- 🟡 **Unconditional application** — now has an explanatory comment noting the graph panel renders on desktop too, so the reader understands the intent. ✓
- 🔵 **Inline style vs. Tailwind** — swapped for `pt-[max(0.75rem,env(safe-area-inset-top,0.75rem))]`, keeping the padding co-located with the other utility classes. ✓

The `py-3` still provides the `padding-bottom` and the arbitrary `pt-[...]` overrides the `padding-top` (Tailwind emits both as separate classes, and the more specific `pt-*` wins via source order). Desktop stays at 0.75rem; iOS gets the safe-area inset. Clean.

Remaining item is just the manual iOS device verification in the test plan before merge.

PR #88, opened 2026-04-20
Fix XBRL concept search shadowed by chart UUID route (#74)

### LINKED ISSUE #74: Search of xbrl concepts when selected as a data source doesn't work
the search doesn't find anything when using income for instance for Shelly Group SE and I know for a fact that such data exists.

nvasilevv:
## Summary
- `/api/charts/xbrl-concepts` was registered after `/api/charts/{chart_id}` in the same router, so FastAPI caught every request with the UUID route first and returned `422 uuid_parsing` before the search handler ran. The frontend treated that as an empty result — the "no concepts found" state from #74.
- Moved the static `xbrl-concepts` route above the `{chart_id}` routes. Added a regression test that resolves the route and asserts the search handler runs.

## Test plan
- [x] `cd backend && python -m pytest tests/test_charts_routes.py -v` — 2 passed (run inside the api container; verified the tests fail on the pre-fix file and pass on the fix).
- [x] Hot-patched the live prod api container and hit `GET /api/charts/xbrl-concepts?ticker=SLYG&keyword=income` — returns 26 concepts incl. `ComprehensiveIncome`, `FinanceIncome`, `IncomeTaxesPaidClassifiedAsOperatingActivities`. Container force-recreated afterwards to restore the image.
- [ ] After merge + deploy: reproduce the original flow in the UI — select "Shelly Group SE" on Charts, pick XBRL Facts, search "income" → results appear.

Closes #74


nvasilevv:
## Summary

PR #88 fixes a real 422-on-search bug by moving the `/xbrl-concepts` handler above `/{chart_id}` in `backend/app/api/charts.py` so FastAPI's order-dependent matcher stops interpreting the literal path segment as a UUID. It adds two route-resolution tests that act as a genuine regression guard: they build a `TestClient` from the production router, so if the ordering ever regresses the tests will 422.

## Review Findings

### 🔴 Critical

- **Duplicate route still present — fix is incomplete.** `backend/app/api/charts.py` now has **two** `@router.get("/xbrl-concepts")` registrations: the moved one at `charts.py:202` and an older duplicate at `charts.py:223` (introduced by #59, never noticed). The PR moved/kept the first copy but didn't delete the second. The fix works because first-match wins in FastAPI, but:
  - The duplicate at line 223 is dead code with a **different signature** (`ticker: str, keyword: str | None = None` — no `Query(...)` metadata, different docstring, no keyword=None normalization). If someone renames or removes the top handler, the second silently takes over with subtly different validation/docs.
  - The "ordering" comment at line 196–201 explains the invariant, but there's no guard preventing a future edit to line 223 from resurrecting the bug under a new name.

  **Action:** delete lines 222–233 before merging.

### 🟡 Improvement

- **Test mock leaks into module state.** `_build_client` does `repository.list_xbrl_concepts_for_ticker = repo_mock` without restoration. If any other test in the same pytest session imports the real function (directly or transitively), it'll get the mock. Switch to `monkeypatch.setattr` via a fixture, or wrap in try/finally. (`test_charts_routes.py:27`)
- **Test doesn't lock in the ordering invariant explicitly.** The tests will catch a full revert, but not a more subtle regression — e.g., someone adding a *new* catch-all route above `/xbrl-concepts`. Consider also asserting `charts_router.routes` ordering: the `/xbrl-concepts` route must appear before any `/{...}` parameterized route. Cheap, robust, documents the invariant in code.

### 🔵 Nitpick

- `args[1]`/`args[2]` positional indexing in the test (`test_charts_routes.py:46-47, 59-60`) — prefer `repo_mock.assert_awaited_once_with(ANY, "SLYG", "income")` or call via kwargs for resilience to signature changes.
- The comment block at `charts.py:196-201` is good, but consider also adding a one-liner at the top of the `/{chart_id}` block (e.g. `# Any new static /charts/<name> routes must be registered ABOVE this line.`) — that's where the next person is most likely to make the mistake.
- Deferred import of `list_xbrl_concepts_for_ticker` inside the handler is inconsistent with neighboring handlers; trivial.

## Positive Feedback

- Correct root-cause diagnosis (route ordering) rather than patching the symptom in the frontend. The header comment explaining *why* the order matters is exactly right.
- The regression test is smart: it instantiates the production `charts_router` so a future revert of the ordering genuinely breaks CI — a real guard, not a tautology.
- The test plan documents an end-to-end verification against the live container before requesting review, which is appropriately thorough for a routing bug that unit tests alone can miss.

## Conclusion

**Request Changes.** The directional fix is correct and the tests are well-scoped, but the duplicate `search_xbrl_concepts` at `charts.py:223` must be removed — shipping with two handlers for the same path is a landmine for the next developer. The test-cleanup item is a should-fix; the ordering-assertion and nitpicks are optional.

nvasilevv:
Thanks for the review. Addressed in 312a0fc:

- **🟡 Test mock leak** — switched to `monkeypatch.setattr` so the stub no longer persists past the test.
- **🟡 Ordering invariant** — added `test_static_xbrl_concepts_route_registered_before_chart_id_parametric_route` that inspects `charts_router.routes` and fails if a `/charts/{...}` parametric route is registered before `/charts/xbrl-concepts`. Verified it fails on pre-fix code and passes on the fix.
- **🔵 Positional args** — replaced `args[1]`/`args[2]` with `assert_awaited_once_with(ANY, "SLYG", ...)`.

On the 🔴 critical finding — I can't reproduce the duplicate. `grep -n '@router\.' backend/app/api/charts.py` on the branch head (312a0fc) shows exactly one `@router.get("/xbrl-concepts")` at line 106, and the containing `async def search_xbrl_concepts` follows immediately. The original duplicate at line 223 of the pre-fix file was the one I moved — I deleted the old location as part of the first commit (2daabde). Could you double-check against the pushed branch rather than a local view that might include my pre-commit state?

Skipped the optional nits (warning comment above `/{chart_id}`, un-deferring the import) per scope — the existing header comment already documents the ordering constraint, and the deferred import matches the pattern used by several neighbors in the same file.

PR #90, opened 2026-04-20
XBRL chart populate: disambiguate period frequency and dedupe comparators

### LINKED ISSUE #89: XBRL chart populate: disambiguate period frequency and dedupe comparator re-reports
## Problem

When populating a chart from `xbrl_facts`, two issues make the resulting series confusing or wrong:

1. **Frequency is collapsed.** `query_xbrl_facts_by_ticker` in `backend/app/services/repository.py:1295-1301` reduces the context to a single `"YYYY-MM-DD to YYYY-MM-DD"` string and drops the raw `start_date`/`end_date`. The chart populate mapping in `backend/app/api/charts.py:299-305` keeps only `period`, `value`, `concept`, `unit`. Consumers can't tell Q / H1 / 9M / FY apart, and a chart that mixes quarterly and annual facts is not comparable on one Y-axis.
2. **Prior-year comparators are duplicated.** Each annual filing re-reports prior years as comparators, so the same `(concept, start_date, end_date, value)` tuple appears in N filings. Example: chart `6e281915-0263-4c76-9103-15a2cda0e870` (SLYG, `Revenue`) has 22 rows, but they collapse to ~5 unique annual values.

## Proposal

**1. Surface period dates and a derived `period_type`.**

In the populate mapping, keep `start_date`, `end_date`, `instant`, and derive `period_type`:

| days (`end - start`) | `period_type` |
|---|---|
| ~90 | `Q` |
| ~180 | `H1` |
| ~270 | `9M` |
| ~365–366 | `FY` |
| `instant` set, no range | `instant` |

Use tolerant bucketing (e.g. 85–95, 175–185, 260–280, 360–370) to absorb calendar drift.

**2. Add an optional `period_type` filter to `PopulateRequest`.**

```python
class PopulateRequest(BaseModel):
    ...
    period_type: Literal["Q", "H1", "9M", "FY", "instant"] | None = None
```

Translated to a WHERE clause on `xbrl_context.end_date - start_date` (or `instant IS NOT NULL`).

**3. Dedupe comparator re-reports.**

After filtering, collapse on `(concept, start_date, end_date)` (or `(concept, instant)` for balance-sheet items), keeping the fact from the latest `filing_date`. This drops the comparator copies while preserving the newest restatement.

## Scope

- `backend/app/services/repository.py` — `query_xbrl_facts_by_ticker`: emit raw dates + `period_type`, accept `period_type` filter.
- `backend/app/api/charts.py` — `PopulateRequest` adds `period_type`; mapping keeps new fields; dedupe before appending to `data_rows`.
- `frontend/src/app/[locale]/charts/page.tsx` — "Data Source" tab exposes a period-type selector when `dsType === "xbrl_facts"`; auto-pick `period_type` as an additional series dimension in previews.

## Out of scope

- Pivoting (one series per `period_type` or `concept`) — the chart editor already lets users re-key after the data is loaded. Revisit if the UX becomes a friction point.

nvasilevv:
Closes #89.

## Summary

When populating a chart from `xbrl_facts`, two issues made the resulting series confusing or wrong:

1. **Frequency was collapsed.** The repository query flattened context into a single `"YYYY-MM-DD to YYYY-MM-DD"` string and the populate mapping dropped everything except `period`, `value`, `concept`, `unit`. A chart mixing quarterly and annual facts couldn't be read on one Y-axis.
2. **Prior-year comparators were duplicated.** Each annual filing re-reports prior years as comparators, so the same `(concept, start, end, value)` tuple appeared in N filings. The example from the issue — chart `6e281915…` (SLYG, `Revenue`) — had 22 rows collapsing to ~5 unique annual values.

This PR fixes both.

### Backend

- **`backend/app/services/repository.py`**
  - New `_derive_period_type(start, end, instant)` helper classifying contexts into `Q` / `H1` / `9M` / `FY` / `instant` using tolerant day-delta buckets (85–95 / 175–185 / 260–280 / 360–370 days) to absorb calendar drift.
  - `query_xbrl_facts_by_ticker` accepts an optional `period_type` parameter that translates to a SQL WHERE on `XbrlContext.end_date - XbrlContext.start_date` (Postgres native int days), or `instant IS NOT NULL`.
  - Each returned row now includes `start_date`, `end_date`, `instant`, and `period_type` alongside the legacy `period` string (kept for backward compat — the helper is also called from agent tooling).

- **`backend/app/api/charts.py`**
  - `PopulateRequest` gains `period_type: Literal["Q","H1","9M","FY","instant"] | None`.
  - The `xbrl_facts` branch forwards `period_type` to the repo, includes the new fields in `data_rows`, and dedupes on `(concept, start, end)` / `(concept, instant)` keeping the row with the newest `filing_date` so restatements win over original reports.
  - Data rows are sorted by period start for a stable chart X-axis.

### Frontend

- **`frontend/src/app/[locale]/charts/page.tsx`** — new `dsPeriodType` state and a `<select>` in the Data Source tab (rendered only for `xbrl_facts`) with options: All / Q / H1 / 9M / FY / Instant. Passed through to the populate call.
- **`frontend/src/lib/api.ts`** — `populateChartData` signature extended with optional `period_type`.

### Tests

- **`backend/tests/test_xbrl_period_type.py`** (new, 12 tests):
  - `_derive_period_type` bucket edges (inclusive 85 → `Q`, exclusive 84 → None; 95 → `Q`, 96 → None), half-year/9M/FY buckets, leap-year tolerance, instant-only, range-wins-over-instant, both-None.
  - Dedupe logic: same `(concept, start, end)` across filings → only the newest `filing_date` kept; distinct periods / concepts / instants all preserved; instants keyed separately from ranges.

## Out of scope

- Pivoting to one series per `period_type` or `concept` — the chart editor already lets users re-key after the data is loaded, and the issue explicitly defers this.

## Test plan

- [x] `python3 -m pytest backend/tests/test_xbrl_period_type.py -v` — 12/12 pass
- [x] Full backend suite (minus a pre-existing `fastapi`-import failure in the local env): 61/61 pass
- [x] `npx tsc --noEmit` on the frontend — clean
- [ ] Manual: populate chart `6e281915-0263-4c76-9103-15a2cda0e870` (SLYG / Revenue) with `period_type=FY` and verify 22 rows collapse to ~5
- [ ] Manual: try each `period_type` option and `All` in the UI dropdown


nvasilevv:
## Review

### Summary

Solid, focused fix. Adds `period_type` classification (Q/H1/9M/FY/instant) to XBRL fact queries, dedupes comparator re-reports by keeping the latest filing, and wires a dropdown through the populate UI. Well-scoped, minimal collateral change.

### Findings

🟡 **Improvement — limit applied before dedupe** (`repository.py:1277`)
`query_xbrl_facts_by_ticker` orders by `desc(filing_date)` then `.limit(limit)`, so duplicates consume the budget *before* the API-layer dedupe runs. For the example in the PR body (22 rows → 5 unique years, `limit=100`) this is fine, but for an issuer with long history and many restatements, older unique years can fall off the limit while recent-year duplicates survive, producing unpredictable X-axis coverage. Consider either (a) dedupe in SQL via `DISTINCT ON (concept, start_date, end_date, instant) ... ORDER BY ..., filing_date DESC`, (b) oversample in the repo call (e.g. `limit * 3`) and truncate after dedupe, or (c) document the interaction so callers raise `limit` appropriately.

🟡 **Improvement — DB-engine assumption** (`repository.py:1311-1313`)
`XbrlContext.end_date - XbrlContext.start_date` evaluating to int days is Postgres-specific. On SQLite it returns a non-integer representation and the `lo <= day_delta <= hi` predicate will misbehave or error. Not a prod bug (we're on Postgres), but the PR adds no test for the SQL path, so a contributor running tests against SQLite would hit this silently. Worth either a comment calling out the Postgres dependency or an integration test.

🟡 **Improvement — dedupe tested by re-implementation, not by import**
`backend/tests/test_xbrl_period_type.py:45-59` duplicates the `charts.py` dedupe logic and tests the copy. If someone changes the real implementation in `charts.py` (key shape, tiebreaker), the tests continue to pass against the stale copy. Extract the dedupe into a small `_dedupe_xbrl_facts` helper and import it from the test — same footprint, real coverage.

🔵 **Nitpick — bucket gaps are silently unclassified**
Days in 96–174, 186–259, 281–359 return `None` → no `period_type` field. Dedupe still works on `(concept, start, end)`, but these facts can't be filtered via the UI. Likely intentional, but a one-line comment on `_PERIOD_TYPE_BUCKETS` noting the gap behavior would help future readers.

🔵 **Nitpick — `_derive_period_type` underscore**
Function is imported from tests. The `_` prefix typically signals "module-private"; since tests explicitly reach for it, drop the underscore or document that tests are the exception.

🔵 **Nitpick — UI label**
`frontend/src/app/[locale]/charts/page.tsx:797-808` renders a bare `<select>` with no label. A short "Period frequency" label would help discoverability.

🔵 **Nitpick — `filing_date` exposed in `data_rows`**
`charts.py` writes `filing_date` onto the row, which then persists into `data_json`. If that's intentional (traceability), fine; otherwise drop it from the output loop — dedupe already consumed it.

### Positive feedback

- Tolerant day-delta buckets with explicit edge tests (`test_quarter_bucket_inclusive_edges`, `test_full_year_bucket_handles_leap`) — right abstraction for calendar drift.
- Keeping the legacy `period` string while additively adding the new fields is correct backward-compat for the agent-tooling caller.
- `Literal[...]` on `period_type` at both the Pydantic boundary and the TS `api.ts` signature closes the loop against invalid inputs reaching the SQL switch.
- Sort-after-dedupe on `start_date || instant || period` gives a stable X-axis regardless of filing order.
- Dedupe key design (`(concept, start or instant, end)`) cleanly separates range and instant facts without a second codepath.

### Security

No concerns. All filters use ORM bindings, `period_type` is Literal-validated, no new PII/secret surface, auth deps unchanged.

### Conclusion

**Comment** — ship-able as-is if the limit-vs-dedupe interaction is acceptable at current data volumes; otherwise **Request Changes** on that one finding. Could you address the three 🟡 items (or explicitly dismiss them with reasoning) before merging? The nits are optional.

nvasilevv:
Thanks for the review — addressed all three 🟡 items and all three 🔵 nits in 688a44d.

**🟡 Limit-vs-dedupe interaction**
Went with the oversample approach (option b): the `xbrl_facts` branch in `charts.py` now requests `min(limit * 5, 500)` from the repo (covers typical 4-prior-year comparator bundles), dedupes, sorts by period start, then truncates to the caller's `limit`. The repo function itself is unchanged so agent-tooling callers keep their old semantics.

**🟡 DB-engine assumption**
Added an explicit Postgres-only comment on the `XbrlContext.end_date - XbrlContext.start_date` filter noting that SQLite returns a string and the predicate would break. No integration test added — no Postgres fixture harness exists yet.

**🟡 Dedupe tested by re-implementation**
Extracted `dedupe_xbrl_facts` to `app.services.repository` (next to `query_xbrl_facts_by_ticker` — semantically XBRL fact post-processing, and avoids the fastapi import chain so tests can load it locally). `charts.py` imports from there; tests now import the real helper. Also renamed `_derive_period_type` → `derive_period_type` since it's a public entry point for tests.

**🔵 Bucket gaps** — added a comment on `_PERIOD_TYPE_BUCKETS` documenting that days in 96–174 / 186–259 / 281–359 classify as `None` and aren't reachable via the UI filter.

**🔵 UI label** — wrapped the select in a `<label>` reading "Period frequency".

**🔵 `filing_date` leak** — dropped it from the row output. Dedupe consumes it internally; no need to persist into `data_json`.

Rechecked: `pytest tests/ --ignore=tests/test_notes_service_concurrency.py` → 61/61 pass (same pre-existing local env issue with fastapi imports). `tsc --noEmit` clean.

